PR #18, opened 2015-02-05
Fix non-ASCII password compatibility with pre-KitKit Android

dwmw2:
The PBKDF2 function in Android used to use only the low 8 bits of each
character. So if we have a non-ASCII (well, non-ISO8859-1) password and
it doesn't work, try mangling it the same way.

See http://forum.yubico.com/viewtopic.php?f=26&t=1601#p6807 for details.


TomAtYubico:
Looks good, thank you for the contribution and spotting this issue for both platform.

PR #20, opened 2015-03-28
Correctly support 8 digit TOTP

lukegb:
This is necessary to allow supporting 8 digit TOTPs (hello, Blizzard Authenticator)


TomAtYubico:
Thank you @lukegb for your contribution

PR #23, opened 2015-05-04
Update `Usage` section to include helpful option for scdaemon

DarkDimius:
It seems that Neo cannot be used by both gpg and Authenticator at the same time.
The problem is more severe, as gpg-agent and scdaemon by default tend to leave connection to card open indefinitely.
This makes opening Authenticator after any usage of gpg fail with error:
```
No Yubikey NEO found. Please plugin your Yubikey NEO in one of your USB port
```

This change advises user to add configuration option that will instruct gpg-agent and scdaemon to close connection to card after 5 seconds of inactivity.


TomAtYubico:
Thank you for your contribution.

PR #41, opened 2015-08-15
Update README

glnds:
Added a note about enabling CCID connection mode in order to properly detect a Yubikey under OS X


dainnilsson:
The information in this PR isn't quite accurate, so I've added a slightly different note to the README which should convey the needed information.

Issue #64, opened 2016-01-11
New credential overwrites current one


pnickerson-cashstar:
I am running yubioath-desktop-2.3.0-1.fc23.noarch and ykpers-1.17.2-3.fc23.x86_64 on Fedora 23.

If I have a current credential present with a given label (user) and issuer, such as "test1" and "testA", and then add a new credential with the same label and issuer but different secret, then the current credential gets overwritten by the new one without warning.

To reproduce, run these two commands in Linux to create two QR codes to scan, one at a time. When scanning the second one, first wait for the 30 second timer to restart and then look at the new current credential's code to see that the new credential overwrites it. I find that I need to reduce the font size of my terminal for the QR code scan to work, or you can use an image --type if you want.
```shell
qrencode --type ANSI256 --margin 1 "otpauth://totp/test1?secret=7777AAAA&issuer=testA"
qrencode --type ANSI256 --margin 1 "otpauth://totp/test1?secret=AAAA7777&issuer=testA"
```

I would expect a warning dialog to appear before the current credential is overwritten by the new one, like the "Confirm credential deletion" dialog that appears when deleting a credential.

dainnilsson:
The credential is uniquely identified by the name (issuer and label, combined), and thus all names must be different (you've probably concluded this already). So the behavior is correct in that one gets overwritten. I agree that the tool should ask the user for confirmation before overwriting an existing credential.

PR #80, opened 2016-05-25
Add confirmation when overwriting entries

dagheyman:
When a entry is added to a Yubikey with an existing name (issuer + label),
ask the user for confirmation to overwrite the existing entry.

Should close issue #64


dainnilsson:
Looks good!

Issue #93, opened 2016-08-26
'Scan a QR code' usually doesn't work


tycho:
The 'Scan a QR code' button rarely seems to work properly for me. I don't know enough about the deep technical details of QR code handling to understand why it's breaking. Below is an example of a QR code that isn't handled correctly.

![qrcode](https://cloud.githubusercontent.com/assets/29616/17999747/7b304e74-6b2f-11e6-94d6-f11644534267.png)

I was able to get yubioath-gui to read arbitrary QR codes reliably by using the 'zbar' library instead of the pure Python QR handling that is currently in use. I posted my changes here: https://github.com/tycho/yubioath-desktop/commit/2af74decba84ae332df3a04dcaefa3bfc706551e

I am not particularly happy with my changes though, which is partly why this isn't a pull request. The additional dependency on zbar doesn't bother me, but the dependency on PIL does. It's a massive library, and I only use it to convert the screenshot image to grayscale in order to make zbar happy.

tycho:
Heh. I just noticed the try/except was eating exceptions and hiding the problem:

```
Traceback (most recent call last):
  File "/home/steven/Development/yubioath-desktop-venv/lib/python2.7/site-packages/yubioath_desktop-3.0.2.dev0-py2.7.egg/yubioath/gui/view/add_cred_legacy.py", line 104, in _do_scan_qr
    data = decode_qr_data(qr)
  File "/home/steven/Development/yubioath-desktop-venv/lib/python2.7/site-packages/yubioath_desktop-3.0.2.dev0-py2.7.egg/yubioath/gui/qrdecode.py", line 64, in decode_qr_data
    data, bits = parse_bits(bits, version)
  File "/home/steven/Development/yubioath-desktop-venv/lib/python2.7/site-packages/yubioath_desktop-3.0.2.dev0-py2.7.egg/yubioath/gui/qrdecode.py", line 267, in parse_bits
    buf += ALPHANUM[num / 45]
TypeError: string indices must be integers, not float
```

This fixes that problem, and makes my example QR code decode properly:

```diff
diff --git a/yubioath/gui/qrdecode.py b/yubioath/gui/qrdecode.py
index c7fab1c..af289ed 100644
--- a/yubioath/gui/qrdecode.py
+++ b/yubioath/gui/qrdecode.py
@@ -263,7 +263,7 @@ def parse_bits(bits, version):
         while l > 0:
             if l >= 2:
[...]

PR #94, opened 2016-08-26
qrdecode: fix error where float was used as a string index

tycho:
We needed to do an integer division not a float division.

This will theoretically fix what was breaking me in issue #93. It at least fixes the test case I provided.


dainnilsson:
Thanks, looks like I missed this when adding support for Python 3!

PR #99, opened 2016-09-11
add auto-completing search text box

tycho:
I've got enough entries in my YubiKey that having a search box on the Authenticator app is kind of useful. So I added one.

Adds two keyboard shortcuts as well:
- Ctrl-F will set the focus to the search text box.
- Escape will clear the contents of search text box.


dainnilsson:
Thanks, that's a really nice feature!

PR #100, opened 2016-09-11
gui controller: unset _reader and _creds on forget_passwords()

tycho:
I was running into this exception when I restore the app from the system tray:

```
    Traceback (most recent call last):
      File "/usr/lib/python2.7/site-packages/yubioath/yubicommon/qt/classes.py", line 86, in customEvent
	event.callback()
      File "/usr/lib/python2.7/site-packages/yubioath/yubicommon/qt/worker.py", line 48, in callback
	self._callback()
      File "/usr/lib/python2.7/site-packages/yubioath/gui/controller.py", line 262, in _init_dev
	dev.unlock(self._keystore.get(dev.id))
      File "/usr/lib/python2.7/site-packages/yubioath/core/standard.py", line 241, in unlock
	resp = self._send(INS_VALIDATE, data)
      File "/usr/lib/python2.7/site-packages/yubioath/core/standard.py", line 179, in _send
	resp, status = self._device.send_apdu(0, ins, p1, p2, data)
      File "/usr/lib/python2.7/site-packages/yubioath/core/ccid.py", line 48, in send_apdu
	resp, sw1, sw2 = self._conn.transmit(header + [byte2int(b) for b in data])
      File "/usr/lib/python2.7/site-packages/smartcard/CardConnectionDecorator.py", line 82, in transmit
	return self.component.transmit(bytes, protocol)
      File "/usr/lib/python2.7/site-packages/smartcard/CardConnection.py", line 146, in transmit
	data, sw1, sw2 = self.doTransmit(bytes, protocol)
      File "/usr/lib/python2.7/site-packages/smartcard/pcsc/PCSCCardConnection.py", line 205, in doTransmit
	SCardGetErrorMessage(hresult))
    CardConnectionException: Failed to transmit with protocol T1. Card was reset.
```

Killing off the existing 'reader' object is sufficient to tell it to reconnect to the reader properly before trying to do anything with it.

### DIFF
--- a/yubioath/core/ccid.py
+++ b/yubioath/core/ccid.py
@@ -31,6 +31,17 @@ from smartcard.Exceptions import SmartcardException
 
 from yubioath.yubicommon.compat import byte2int, int2byte
 
+__all__ = ['ScardDevice', 'open_scard']
+
+# Mimics the weakref object interface while using strong references. Useful for
+# places where you actually need a strong reference but the interface you're
+# using only supports weakrefs.
+class strongref(object):
+    def __init__(self, value):
+        self._value = value
+
+    def __call__(self):
+        return self._value
 
 class ScardDevice(object):
 
@@ -45,12 +56,14 @@ class ScardDevice(object):
         header = [cl, ins, p1, p2, len(data)]
         # from binascii import b2a_hex
         # print("SEND:", b2a_hex(''.join(map(int2byte, header)) + data))
-        resp, sw1, sw2 = self._conn.transmit(header + [byte2int(b) for b in data])
+        resp, sw1, sw2 = self._conn().transmit(header + [byte2int(b) for b in data])
         # print("RECV:", b2a_hex(b''.join(map(int2byte, resp + [sw1, sw2]))))
         return b''.join(int2byte(i) for i in resp), sw1 << 8 | sw2
 
     def close(self):
-        self._conn.disconnect()
+        conn = self._conn()
+        if conn:
+            self._conn().disconnect()
 
     def __del__(self):
         self.close()
@@ -63,6 +76,6 @@ def open_scard(name='Yubikey'):
             conn = reader.createConnection()
             try:
                 conn.connect()
-                return ScardDevice(conn)
+                return ScardDevice(strongref(conn))
             except SmartcardException:
                 pass
--- a/yubioath/gui/ccid.py
+++ b/yubioath/gui/ccid.py
@@ -82,6 +82,7 @@ class CardWatcher(QtCore.QObject):
         self._status = CardStatus.NoCard
         self.reader_name = reader_name
         self._callback = callback or (lambda _: _)
+        self._conn = None
         self._reader = None
         self._reader_observer = _CcidReaderObserver(self)
         self._card_observer = _CcidCardObserver(self)
@@ -122,13 +123,18 @@ class CardWatcher(QtCore.QObject):
 
     def open(self):
         if self._reader:
-            conn = self._reader.createConnection()
+            if self._conn is None:
+                self._conn = self._reader.createConnection()
             try:
-                conn.connect()
+                self._conn.connect()
                 self._set_status(CardStatus.Present)
-                return ScardDevice(conn)
+                return ScardDevice(weakref.ref(self._conn))
             except SmartcardException:
                 self._set_status(CardStatus.InUse)
+
+    def close(self):
+        del self._conn
+        self._conn = None
 
     def __del__(self):
         self._reader_observer.delete()
--- a/yubioath/gui/controller.py
+++ b/yubioath/gui/controller.py
@@ -417,4 +417,5 @@ class GuiController(QtCore.QObject, Controller):
     def forget_passwords(self):
         self._reader = None
         self._creds = None
+        self.watcher.close()
         self._keystore.forget()
--- a/yubioath/gui/controller.py
+++ b/yubioath/gui/controller.py
@@ -415,4 +415,6 @@ class GuiController(QtCore.QObject, Controller):
                 self._keystore.put(dev.id, key, remember)
 
     def forget_passwords(self):
+        self._reader = None
+        self._creds = None
         self._keystore.forget()


tycho:
Also added a fix to release the smart card when minimizing/closing to the system tray. Otherwise the card is still tied up by the authenticator despite not being *used* by it at all.

dagheyman:
Just want to say a quick thank you for your contributions! I hope we may be able to review them soon.

tycho:
@dagheyman Happy to help out, this is a fun little project. My contributions are totally selfish though: they make my daily life easier because I use the Yubico Authenticator all the time. :wink:

dainnilsson:
@tycho We're very happy for the contributions! Can I get you to state that the code you submit to this project is your own, and that you agree to contribute it under the license of the project (GPLv3)?

We have some other high-priority stuff going on these coming two weeks, so please bear with us if we don't have time to look at your submissions until afterwards.

tycho:
@dainnilsson Of course it's my own code. And although I personally dislike the GPL license, I agree to contribute under your project's license ("when in Rome..."). Ordinarily I just license everything I write under the 2-clause BSD license, which is [GPL-compatible](https://www.gnu.org/licenses/license-list.en.html#FreeBSD).

dainnilsson:
@tycho I had no doubt that it was! We generally prefer BSD-2-clause as well. This project is GPL mainly for historical reasons I believe...

dainnilsson:
With a password protected (not "remembered") YubiKey I'm getting an Exception a lot after entering the password on start. This is on Ubuntu and running this PR merged with master. Unfortunately I don't have time to investigate the issue right now.

```
Traceback (most recent call last):
  File "/home/dain/yubico/py/yubioath-desktop/yubioath/yubicommon/qt/classes.py", line 86, in customEvent
    event.callback()
  File "/home/dain/yubico/py/yubioath-desktop/yubioath/yubicommon/qt/worker.py", line 48, in callback
    self._callback()
  File "/home/dain/yubico/py/yubioath-desktop/yubioath/yubicommon/qt/worker.py", line 105, in callback
    def callback(e): raise e
CardConnectionException: Card not connected
```

tycho:
@dainnilsson That's odd. I've been using this for almost two weeks on Mac OS X, Arch Linux, and Windows without any such issues.

tycho:
@dainnilsson Just tried it again, still no issues even with all the merges done on upstream. I even wiped a machine, installed Ubuntu 16.04 and tried it there. Still no repro for the issue you ran into. :disappointed:

dainnilsson:
This is how I'm reproducing it:
1. Have a YubiKey 4 (4.3.1 in my case, I've seen it happen on a NEO as well) with a password set but not remembered, and at least one credential.
2. Launch the app
3. When prompted, enter the password.

It doesn't happen every time, but once in a while it fails with the following trace:

```
$ yubioath-gui 
Traceback (most recent call last):
  File "/home/dain/yubico/py/yubioath-desktop/yubioath/yubicommon/qt/classes.py", line 86, in customEvent
    event.callback()
  File "/home/dain/yubico/py/yubioath-desktop/yubioath/yubicommon/qt/worker.py", line 48, in callback
    self._callback()
  File "/home/dain/yubico/py/yubioath-desktop/yubioath/yubicommon/qt/worker.py", line 105, in callback
    def callback(e): raise e
  File "/home/dain/yubico/py/yubioath-desktop/yubioath/yubicommon/qt/worker.py", line 99, in work
    result = fn()
  File "/home/dain/yubico/py/yubioath-desktop/yubioath/gui/controller.py", line 339, in _refresh_codes_locked
    False)
  File "/home/dain/yubico/py/yubioath-desktop/yubioath/core/controller.py", line 103, in read_creds
    self.unlock(std)
  File "/home/dain/yubico/py/yubioath-desktop/yubioath/gui/controller.py", line 195, in unlock
    dev.unlock(key)
  File "/home/dain/yubico/py/yubioath-desktop/yubioath/core/standard.py", line 241, in unlock
    resp = self._send(INS_VALIDATE, data)
  File "/home/dain/yubico/py/yubioath-desktop/yubioath/core/standard.py", line 179, in _send
    resp, status = self._device.send_apdu(0, ins, p1, p2, data)
  File "/home/dain/yubico/py/yubioath-desktop/yubioath/core/ccid.py", line 59, in send_apdu
    resp, sw1, sw2 = self._conn().transmit(header + [byte2int(b) for b in data])
  File "/home/dain/.virtualenvs/yubioath/lib/python3.5/site-packages/smartcard/CardConnectionDecorator.py", line 82, in transmit
    return self.component.transmit(bytes, protocol)
  File "/home/dain/.virtualenvs/yubioath/lib/python3.5/site-packages/smartcard/CardConnection.py", line 146, in transmit
    data, sw1, sw2 = self.doTransmit(bytes, protocol)
  File "/home/dain/.virtualenvs/yubioath/lib/python3.5/site-packages/smartcard/pcsc/PCSCCardConnection.py", line 198, in doTransmit
    raise CardConnectionException('Card not connected')
smartcard.Exceptions.CardConnectionException: Card not connected
```

I'm going to keep digging a bit, but wanted to post more info right away.

dainnilsson:
Unfortunately I didn't get much time to look at this today, but I did figure a few things out. There seems to be a race when OK-ing the password prompt. At that time the ScardDevice is GC'ed, so the connection is closed. At the same time, CardWatcher.open() is called to get a new connection. However, the check for "self._conn is None" sometimes happens before the GC in which case no new connection is opened, and the exception is thrown.

dainnilsson:
I've tried to rework some of the code so that the card is released whenever it is not in use in general in a new branch: https://github.com/Yubico/yubioath-desktop/tree/release-card-when-idle. @tycho Can you try that out and see if it resolves the same issues that this PR addresses?

tycho:
@dainnilsson Your change does work as described. But I think maybe it's a little bit *too* fine-grained.

Right now the behaviour (with my change at least) is I think how it was originally designed to be: while the Authenticator window is visible, the Authenticator has exclusive access to the card. Things like `gpg-connect-agent learn /bye` will fail because the Authenticator owns it. In order to allow other apps to have access, I need to close the Authenticator or minimize it to the tray. The contract there is that when the window is not visible, the card is released. The only time it attempts to kill scdaemon is when the window becomes visible (either on initial launch or when restored from the tray).

With your change the contract is different, and might present a general usability issue. If the card gets opened by something else while the window is still visible, it fails to refresh the codes on the next attempt and then waits for the card to be released. If I want it to invoke the "kill scdaemon on show" code path then I need to fire a window "show" event somehow (either by restoring from the tray or re-opening the app). Alternatively, I need to hunt down whatever has exclusive access to the card and kick it myself (e.g. `gpg-connect-agent 'scd killscd' 'scd bye' /bye`) to release the card from gpg-agent.

I've noticed that some things that don't really -need- the card will cause scdaemon to spawn and open the card unnecessarily. I think Kleopatra does this sometimes. It would be frustrating if this happened while the Authenticator is still visible, because then I have to do a dance of minimizing/restoring the Authenticator from the tray.

I'm a bit torn about this. In an ideal world I wouldn't need to micromanage which apps have the card open -- which is one reason I like the change. But since it's annoying to have to hunt down `scdaemon` myself, the contract of "while the window is visible, the card is open" is useful.

Thoughts?

dainnilsson:
@tycho Thanks for the extra thinking around this, I'm pretty much in full agreement with your conclusions. I think keeping the connection open while the app is displaying codes might be a required change to prevent other "stuff" from stealing it. I'll take another stab at it tomorrow.

dainnilsson:
Alright, I've pushed a new commit. Care to give it a try?

tycho:
@dainnilsson Just tested that one, works much better! Thanks for working on this. :smile:

dainnilsson:
Great, I've merged those changes onto master and will now close this issue.

PR #103, opened 2016-09-13
yubioath-gui: fix UI hitches on codes refresh

tycho:
This prevents the user interface from hitching when refreshing the codes list. The timer in the controller was invoking the refresh_codes callback in the UI thread, which caused everything to hitch on refresh.


dainnilsson:
Thanks!

PR #117, opened 2017-02-07
Allow 7 digits

decklin:
I would like to generate TOTP codes for a service that uses 7 digits. This adds that possibility, and moves both the default # of digits and allowed set (now [6, 7, 8]) to constants, which are moved to their own file to avoid a circular import problem.

I've tried this on a YubiKey 4 Nano, and the YubiKey has no problem storing the configuration, and in fact an unpatched version of the `yubioath` command will now output 7 digits for my test credential; the number-of-digits validation happens at the time you run `put`, so this makes sense.

### DIFF
--- a/yubioath/cli/__main__.py
+++ b/yubioath/cli/__main__.py
@@ -30,7 +30,7 @@ from __future__ import print_function
 
 from .. import __version__
 from ..core.ccid import open_scard
-from ..core.standard import ALG_SHA1, ALG_SHA256, TYPE_HOTP, TYPE_TOTP
+from ..core.standard import DEFAULT_DIGITS, ALLOWED_DIGITS, ALG_SHA1, ALG_SHA256, TYPE_HOTP, TYPE_TOTP
 from ..core.utils import parse_uri
 from ..core.exc import NoSpaceError
 from .keystore import get_keystore
@@ -138,7 +138,7 @@ def show(ctx, query, slot1, slot2, timestamp):
 @click.option(
     '-A', '--oath-type', type=click.Choice(['totp', 'hotp']), default='totp',
     help='Specify whether this is a time or counter-based OATH credential.')
-@click.option('-D', '--digits', type=click.Choice(['6', '8']), default='6',
+@click.option('-D', '--digits', type=click.Choice([str(d) for d in ALLOWED_DIGITS]), default=str(DEFAULT_DIGITS),
               callback=lambda c, p, v: int(v), help='Number of digits.')
 @click.option(
     '-H', '--hmac-algorithm', type=click.Choice(['SHA1', 'SHA256']),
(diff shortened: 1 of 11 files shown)


decklin:
Unfortunately, it looks like this fails the flake8 check. Any suggestions? Literally listing every constant in the standard.py `import from`?

dagheyman:
First, thanks for your contribution!

I'm not sure I would want to merge this as it is, since the project is undergoing som structural changes at the moment. The CLI part have been ported to the [YubiKey Manager](https://github.com/Yubico/yubikey-manager/) tool, (`ykman oath -h` for a list of the commands). The GUI is currently being rewritten in QML, and will not be using the code based on PySide anymore.

May I ask what service that is using 7 digits TOTP credentials? It doesn't follow the format described [here](https://github.com/google/google-authenticator/wiki/Key-Uri-Format).

decklin:
OK, thanks, I didn't know I was using the wrong thing, I'll take a look at that one.

The need for 7 digits is for Authy. There are some other peculiarities with their setup, but this seemed like a good place to start.

dagheyman:
So, according to [this](https://www.authy.com/product/options/#softtoken) Authy also uses a 20 seconds time interval. This is currently not supported neither, so allowing 7 digits would not be enough here.

decklin:
That's correct. I have a working, hacky option to change the time interval which I will attempt to clean up and port to `yubikey-manager`.

Would you be willing to accept a `Authy:` prefix on key names the way you do with `Steam:`? That could make things somewhat simpler.

dagheyman:
After some discussion we're planning to add support for different time intervals as well as for 7 digits. We're not sure of the exact format yet, but it will probably be a part of the issuer prefix in some way, maybe `Authy{20}:`. Unusual time intervals has been requested [before](https://github.com/Yubico/yubioath-desktop/issues/90). We also need to figure out the best way to visualise this in the GUI (multiple progress bars?), which as mentioned is currently being rewritten.

dagheyman:
@decklin Btw, how do you extract the secret from the Authy service? When I experimented a little I never seemed to be showed a QR code or similar.

decklin:
@dagheyman it's totally not supported by them, which I'm willing to deal with, but you can take the Chrome desktop app and list the secrets with a bit of JS: https://gist.github.com/decklin/7ae721e4f2c02b03d10b22571885dbc7 (I based this on prior work in the other Gists linked there). I have successfully turned this data into QR codes and scanned it into GA and a few other phone apps.

dagheyman:
Just a FYI, yubikey-manager now has 7 digits and non-30 period support on the master branch. To add a custom period, prefix the credential with [period]/Issuer:name. So for a 20 second period it would be `20/Issuer:username`.

Issue #150, opened 2017-04-05
double-clicking tray icon doesn't raise window


tycho:
The window doesn't appear when the tray icon is double clicked. Instead, I have to right click the icon and click "Show credentials"

zypA13510:
Hi, I have a suggestion here. Now that we can minimize to tray and keep running in the background, why don't we make it auto-hide when the key is unplugged, and auto pop out when the key is plugged in? For me, I think this will be very useful.

dagheyman:
@zypA13510 Hi, thanks for your suggestion. I broke it out to a separate issue.

dagheyman:
This is now fixed on master for Windows and Linux. For macOS, since a context menu is used, [listening on double-click](https://bugreports.qt.io/browse/QTBUG-5770?focusedCommentId=351898&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-351898) is not supported. The fix will be included in the next release.

Issue #149, opened 2017-04-04
suggested ux improvements


tycho:
A few things i noticed playing with the current git version (4.0.1-16-ge0e59dd).

* The File>Exit menu item should either be "E&xit" or "&Quit" (i.e. the keyboard shortcut should be x or q, as with any other app, but not "&Exit" as it currently is).
* The "enter" key does nothing in the passphrase dialog when the checkbox has focus (i.e. enter passphrase, tab to checkbox, space to check it, pressing enter *should* click OK as it does when the text box has focus, but it does not)
* When the search text box has focus, I'd expect the Escape key to clear the current search terms (muscle memory from using search in web browsers like Chrome or Firefox)

tycho:
Another one, a bit more trivial:
* I would suggest that the timer tick rate for updating the progress bar should probably be lower. You'd be surprised how much battery life is impacted in a laptop or tablet when something as simple as a blinking cursor is on screen (it prevents the Intel graphics chip from entering RC6p) -- or in this case, a very rapidly updating progress bar. You could probably update it every second or so and still show just as effectively how much time is remaining.

dagheyman:
Thanks for the feedback, and I agree with all your first three points. 

Regarding the timer, yes it does check every [100 milliseconds](https://github.com/Yubico/yubioath-desktop/blob/master/qml/main.qml#L316) now, maybe it is  possible to check at a lower rate and still have a smooth progress bar (maybe some animations could help?).

tycho:
@dagheyman I'm suggesting that even 100ms is too short a delay if you care about battery life. It keeps the GPU awake to redraw the progress bar. Also, I just noticed the timer doesn't stop ticking when minimized to the system tray, which means it's always consuming CPU even when backgrounded.

drohm:
@dagheyman If you were to implement the change regarding the timer, would that affect desktops?  If so, maybe make it an option that can be enabled in Settings for laptops and such?

dagheyman:
Closing since two of these are fixed and the third one (clear search box) has a separate issue.

Issue #129, opened 2017-03-21
Progress bar color doesn't follow color scheme.


dagheyman:
Maybe it should?

Issue #122, opened 2017-03-20
ux improvement - scroll position resets with timer


xortim:
The scrolling position resets when the timer expires. This can be frustrating for users that are waiting for the next code and have the desired account drawn below the fold.

dagheyman:
Should be solved by b53a1d548cf0afdc50db6446ddfdbd5857ba5fd9.

PR #151, opened 2017-04-05
some UX fixes

tycho:
This fixes some of the UX issues raised in #149.

Issues mentioned there that **ARE** fixed by this pull request:
* The File>Exit menu item should either be "E&xit" or "&Quit" (i.e. the keyboard shortcut should be x or q, as with any other app, but not "&Exit" as it currently is).
* When the search text box has focus, I'd expect the Escape key to clear the current search terms (muscle memory from using search in web browsers like Chrome or Firefox)

Other issues fixed in this pull request:
* The timers are suspended when the window is not visible (i.e. minimized to system tray). They resume as soon as the window is opened again.
* Credentials were being refreshed one second *after* the deadline instead of *on* the deadline.
* Progress bar color now is pulled from SystemPalette.highlight

Issues mentioned there that **ARE NOT** fixed by this pull request:
* The "enter" key does nothing in the passphrase dialog when the checkbox has focus (i.e. enter passphrase, tab to checkbox, space to check it, pressing enter should click OK as it does when the text box has focus, but it does not)

Issues discovered during creation of this pull rquest:
* Double-clicking the tray icon doesn't show the window (#150)

(Amended: I changed my mind about the progress bar update timer tick. As long as the timer is stopped while the app is in the tray, it shouldn't have any appreciable impact on battery life.)

### DIFF
--- a/qml/MainMenuBar.qml
+++ b/qml/MainMenuBar.qml
@@ -38,7 +38,7 @@ MenuBar {
             onTriggered: openSettings()
         }
         MenuItem {
-            text: qsTr("\&Exit")
+            text: qsTr("E\&xit")
             onTriggered: Qt.quit()
             shortcut: StandardKey.Quit
         }
--- a/qml/PasswordPrompt.qml
+++ b/qml/PasswordPrompt.qml
@@ -48,6 +48,7 @@ DefaultDialog {
             Button {
                 id: okBtn
                 text: qsTr("Ok")
+                isDefault: true
                 Layout.alignment: Qt.AlignRight | Qt.AlignBottom
                 onClicked: promptAccepted()
                 KeyNavigation.tab: cancelBtn
--- a/qml/YubiKey.qml
+++ b/qml/YubiKey.qml
@@ -87,7 +87,7 @@ Python {
 
     function refreshCCIDCredentials(force) {
         var now = Math.floor(Date.now() / 1000)
-        if (force || (validated && nextRefresh < now)) {
+        if (force || (validated && nextRefresh <= now)) {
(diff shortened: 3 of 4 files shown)


dagheyman:
Thanks for your contributions.

a814d50 seems ready to merge.

8ecb234 has some issues, I think we only want to capture the keypress when the search field has focus? Also, the selection of the credential is not dropped when esc is pressed.

e831539 seems to work, but maybe we can bind the running property of the timer to the `canShowCredentials` boolean?

7eb1d35 and e4eff70 I think are mostly okay (but need more testing) except I think they break the scroll position? See (#122). I'm unsure about how exactly but the app built with your PR seems to re-introduces this behaviour (reseting scroll position on reset).

9c78cea I thought a bit more about colors, and I think we may want to use the `palette.highlight` color for selected credentials? Maybe the most native-look is to just remove all custom styling for the ProgressBar. Need to test that this looks okay on all platforms though. Lets discuss in #129.

If you want to keep working on these issues in this PR you are welcome, I could also cherry-pick a814d50 to master if you prefer. 

In the future I would prefer separate Issues/PRs for all proposed changes, makes stuff easier to reason about.

tycho:
@dagheyman Separate PRs would be easier to deal with if there were fewer issues. :wink:

I'm on vacation right now so turnaround time will be slower, but I'll try to take a look at your comments tomorrow and try to address them.

tycho:
I rebased these changes. Haven't made any modifications to them yet. I may get to it tomorrow.

dagheyman:
I cherry-picked some of these changes to master, if you are still interested in getting more of them in, feel free to open new PRs (one per issue is preferred).

PR #163, opened 2017-05-20
qrdecode: fix infinite loop when encoding type is 1 (numeric)

tycho:
Found a bug in the QR decoder when doing code coverage testing of my PyQt5 port for v3.x. I managed to find a QR code that got the interpreter into an infinite loop.

See the QR code near the bottom of this page, under the heading "Supported Barcode Types": https://packagist.org/packages/yellowskies/qr-code-bundle


dagheyman:
Nice find, merging!

PR #195, opened 2017-10-20
Add option to hide window on launch

sstallion:
I typically launch yubioath-desktop when my window manager starts. Something I've always wanted to have was for the application to launch to the system tray. I was able to hack something up that works, though there are some side-effects I don't understand (I'm not very savvy when it comes to Qt).

Out of curiosity, is this something that the team might be interested in? If so, I have questions! These changes work with the exception of a bug when leaving the settings dialog if both `Show in system tray` and `Hide on launched` options are unchecked - the main window disappears. I suspect this is due to the change I made to the `visible` property, however I'm not sure of how to make this dynamic. Ideally we would inhibit showing the main window immediately after launch rather than when the option is checked.

Thanks in advance!

Obligatory screenshot below.
screenshot

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -18,3 +18,4 @@ moc_*.cpp
 moc_*.h
 yubioath-desktop.app/
 pymodules
+.vagrant/


emlun:
Moved to #197 by @sstallion.

PR #196, opened 2017-10-20
Apply alternateBase color to progress bar background

sstallion:
The background color for the progress bar is black on fedora, which makes it difficult to read time remaining:
screenshot from 2017-10-20 16-55-52
This PR sets the progress bar background to the alternate base color, which improves readability significantly without requiring a color be selected for the foreground:
screenshot from 2017-10-20 16-55-11

### DIFF
--- a/qml/TimeLeftBar.qml
+++ b/qml/TimeLeftBar.qml
@@ -12,4 +12,9 @@ ProgressBar {
     maximumValue: 30
     minimumValue: 0
     rotation: 180
+    style: ProgressBarStyle {
+        background: Rectangle {
+            color: palette.alternateBase
+        }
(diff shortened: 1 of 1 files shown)


emlun:
Unfortunately this has the side effect of messing up the style of the rest of the progress bar, so that it apparently no longer uses the system palette colors - on your branch the progress bar becomes blue despite my palette in `qt5ct` doesn't contain any blue.

I'm not willing to break theming support for this; sorry, you'll have to tweak your own theme instead. The colors you want to change have the following names in `qt5ct`:

- "Normal background (Active)" for the progress bar background
- "Highlight (Active)" for the progress bar filling

PR #197, opened 2017-10-23
Add option to hide window on launch

sstallion:
I typically launch yubioath-desktop when my window manager starts. Something I've always wanted to have was for the application to launch to the system tray. I was able to hack something up that works, though there are some side-effects I don't understand (I'm not very savvy when it comes to Qt).

Out of curiosity, is this something that the team might be interested in? If so, I have questions! These changes work with the exception of a bug when leaving the settings dialog if both Show in system tray and Hide on launched options are unchecked - the main window disappears. I suspect this is due to the change I made to the visible property, however I'm not sure of how to make this dynamic. Ideally we would inhibit showing the main window immediately after launch rather than when the option is checked.

Thanks in advance!

Obligatory screenshot below.
31841929-582d1e3e-b5b1-11e7-968e-5f3aefbee19f

(Apologies for resubmitting the PR, I accidentally created this off of the master branch)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -17,5 +17,6 @@ qrc_*.cpp
 moc_*.cpp
 moc_*.h
 yubioath-desktop.app/
+yubioath-desktop
 pymodules
 .vagrant/
--- a/main.cpp
+++ b/main.cpp
@@ -94,6 +94,11 @@ int main(int argc, char *argv[])
 
     // Set icon in the window, doesn't effect desktop icons.
     qmlWindow->setIcon(QIcon(path_prefix + "/images/windowicon.png"));
+    // Show root window unless explicitly hidden in settings.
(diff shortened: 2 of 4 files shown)


emlun:
Thank you; at a glance this seems like a feature we would happily accept, but I'm going to need some time to look at this and there's a lot going on right now. I'll get back to you as soon as I can.

sstallion:
Sounds good! I'll do some work in the meantime to see if I can better work
out what's going on and apply a fix. Thanks for taking a look!

sstallion:
@emlun I've just pushed an update. I believe this has the correct behavior. Please let me know if there is anything I can do to help!

emlun:
Looks good, thanks a lot!

sstallion:
My pleasure. Thanks!

Issue #187, opened 2017-09-25
Simplify keyboard usage


minisu:
I love Yubico Authenticator and use it every day. Here's a suggestion for how I think it could be even better:

---

My list of credentials includes the following two:

- Google: my.name@mycompany.com
- my.name@mycompany.com

Currently, in order to paste the value of the second credential I have to:

1. Open/maximize Yubico Authenticator.
1. Press _Tab_ to put focus on the _Search_ text field.
1. Type something, e.g. `mycom`. This filters the list to only the above mentioned credentials.
1. Press _Tab_ to put focus on the list of credentials.
1. Press _Down Arrow_ to select the second credential.
1. Press _Enter_.
1. Minimize Yubico Authenticator.

I propose a few small changes in Yubico Authenticator behavior:

- Search field being focused by default.
- Up/down arrows changing list entries even when the search field is focused.
- Enter key not only copying the OTP to the clipboard, but also minimizing Yubico Authenticator. Ideally combined with a message _"OTP copied to clipboard"_ (I'm thinking something like [Toasts](https://developer.android.com/guide/topics/ui/notifiers/toasts.html) in Android).

With these changes, the above flow would be reduced to:

1. Open/maximize Yubico Authenticator.
1. Type something, e.g. `mycom`. This filters the list to only the above mentioned credentials.
1. Press _Down Arrow_ to select the second credential.
1. Press _Enter_.

This would also harmonize the behavior of Yubico Authenticator with most filterable lists, e.g. the address field in web browsers, search engine suggestions and the start menu in operating systems.

emlun:
Suggestions 1 and 2 are now implemented in master. I'm not sure 3 (minimize on Enter) is something we want to do, I'll need to confer with @dagheyman when he returns from his vacation. Moving that suggestion to #214.

emlun:
By the way: If you want to squeeze keystrokes down to the absolute minimum, and you're on Linux/Unix, you could also try [yubikey-oath-dmenu][yod]. Note however that this is not officially supported by Yubico.

[yod]: https://github.com/emlun/yubikey-oath-dmenu

minisu:
Cool @emlun :ice_cream:

PR #201, opened 2017-11-07
Partially addresses #187

artburkart:
**WIP:**
This PR doesn't address the whole issue, but it does add the up/down
navigation when the cursor is in the search bar. I don't know anything
about qml, so this is probably not clean by any means, but I'm happy to
change it. Maybe I'll learn more about how to write this stuff the
proper way as I continue to chip away at the requests outlined in #187.

### DIFF
--- a/qml/main.qml
+++ b/qml/main.qml
@@ -25,6 +25,12 @@ ApplicationWindow {
     property var hotpCoolDowns: []
 
     property var selected: null
+    property var firstCred: repeater.model[0]
+    property int nCreds: repeater.model.length
+    property var lastCred: repeater.model[nCreds - 1]
+    property bool lastCredSelected: selectedIndex === nCreds - 1
+    property bool firstCredSeleced: selectedIndex === 0
+    property bool nothingSelected: selectedIndex === null
     property var selectedIndex: null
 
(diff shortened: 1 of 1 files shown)


artburkart:
Is the flake8 failure legitimate? I haven't touched any of those files.

emlun:
Thanks for contributing! I'll take a look soon, hopefully. Don't worry about the flake8 failures, that looks like existing lint that's been exposed by a recent upstream configuration change.

artburkart:
@emlun - Thanks for following up, please don't spend too much time on this. I don't know qt or qml, so I have definitely done it wrong. The functionality I'm pondering is whether it would be useful to just search for creds when any letters are typed. I constantly want that while using this tool.

emlun:
Thanks for contributing! Unfortunately you were right in this not being very clean. :) I started out from your solution but ended up with something completely different in the end: c8ab5b7b58bb319f00763f1be9533684641f35ed . Still, you're more than welcome to contribute more PRs if you feel like it!

artburkart:
@emlun, thanks! I figured it would be a learning opportunity for me more than anything. A bit selfish on my end 😛

PR #207, opened 2017-11-16
unbundle qtsingleapplication

blshkv:
See the following bug report for more details:
https://github.com/Yubico/yubioath-desktop/issues/206


emlun:
Thank you, but for now we'll go with bundling the dependencies in the source tarball instead.

Issue #222, opened 2017-12-19
recently CTRL-C broke, only ENTER copies OTP


alagopus:
- Yubico Authenticator version: 4.3.1
- Operating system and version: elementary OS Loki, 4.4.0-104-generic #127-Ubuntu
- YubiKey model and version: Both neo and YK 4
- Bug description summary: Despite the context menu claiming CTRL-C, copy is at ENTER

Attach key, open Yubico Authenticator. Select an Entry, and press CTRL-C. Try pasting: nothing.

Expected result: Current OTP copied to clipboard.

Actual results: No change in clipboard contents.

emlun:
Thanks for reporting! The cause of this is that focus now starts on the search field, so <kbd>Ctrl</kbd>+<kbd>C</kbd> copies from the search field instead of from the selected credential unless you press <kbd>Tab</kbd> to move focus.

We should probably have a little think about this and maybe redesign the keyboard interaction a little bit. Do you prefer copying with <kbd>Ctrl</kbd>+<kbd>C</kbd> over <kbd>Enter</kbd>, or is the issue rather that the hints in the menus are misleading?

alagopus:
First, yes I think I prefer `Ctrl+C` to `Enter`, but I may get used to `Enter` after a while. But why doesn't the entry get focus when selected by mouse or cursor keys? Also, I think it would be fair to change the menu hint to `Enter`, if that's the only key binding now...

alagopus:
There is even more gone wrong with the entry copy. Previously, the selection was also set along with the cutbuffer.  This distinction is a bit X11 specific but has great bearing on useability. The selection is more ephemeral and can be pasted by `Mouse-Middle` button.

Essentially, in most terminals there is no quick access to clipboard paste because `Ctrl-C` is terminal interrupt instead. The recent regression breaks my workflow and I'm investigating ways to roll back or switch to another device.

rmenessec:
Confirmed this is a problem. I seem to recall we had an identical or very similar bug open for this behavior, and it was previously fixed months ago.

The shortcut listed in the menu is Ctrl+C; therefore the shortcut should be Ctrl+C, and it should work at all times unless the UI/UX clearly suggests that there should be an exception.

[...]

Issue #226, opened 2017-12-27
"Generate code" shortcut doesn't work


rettlsteiner:
The "Generate code" shortcut doesn't work, the Authenticator will search for " " (space).

Versions:
Yubico Authenticator v4.3.2
YubiKey OATH v4.3.7

dagheyman:
Thanks for the report! It's related to #222 , and will be addressed in the next version.

dagheyman:
Fixed on master, will be in next release. The search field will continue to have the default focus, but the space shortcut will work when if the credential-list has focus. Pressing Enter (generate and copy) will work from "within" the search.

PR #227, opened 2018-01-02
Fix keyboard shortcuts from search 

dagheyman:
Related to #222 .

These changes still has the search field as default focus, but allows CTRL/CMD + C to be used to copy the selected credential and re-enables ESC for clearing the search field.

### DIFF
--- a/qml/main.qml
+++ b/qml/main.qml
@@ -329,14 +329,16 @@ ApplicationWindow {
                 onActivated: search.forceActiveFocus()
             }
             onTextChanged: selectFirstSearchResult()
-            Keys.onEscapePressed: {
-                search.clear()
-                deselectCredential()
-            }
+            Keys.onEscapePressed: search.clear()
             Keys.onReturnPressed: generateOrCopy()
             Keys.onEnterPressed: generateOrCopy()
             Keys.onDownPressed: arrowKeys.goDown()
             Keys.onUpPressed: arrowKeys.goUp()
+            function clear() {
+                search.text = ""
+                arrowKeys.forceActiveFocus()
+                deselectCredential()
+            }
         }
     }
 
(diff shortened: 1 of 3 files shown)


qml/main.qml
```diff
@@ -326,17 +329,36 @@ ApplicationWindow {
...
+            Keys.onPressed: {
+                if(event.matches(StandardKey.Copy)) {
+                    copyAction.trigger()
+                }
+            }
+
+            function clear() {
+                search.text = ""
+                arrowKeys.forceActiveFocus()
+                deselectCredential()
```
emlun: Sure about clearing the selection? It's now been fixed to be stable across changing search queries; I feel like it makes more sense to keep the selection where it is (by just deleting this line).

emlun:
This solves #222, but not #226. Should we remove the <kbd>Space</kbd> shortcut for generating codes?

qml/main.qml
dagheyman: Hmm, really? If I'm clearing the search with the ESC key, I wouldn't want any credential to be selected. I think this pattern is pretty common in search fields.

dagheyman:
> Should we remove the Space shortcut for generating codes?

I don't know, I thought about it as well. Right now it works when the search field doesn't have focus. Do you mean removing just the keyboard-shortcut or removing the action (generate but don't copy) completely?

My guess would be that there is some users that wants to be able to generate codes without adding the OTP to the clipboard.

emlun:
Oh right, it's a different action. I meant just removing the keyboard shortcut. In that case perhaps we should revert to not focusing the search field by default.

qml/main.qml
emlun: Hm, ok. I think I'd expect that if I type "github" into the search box and get my GitHub credential selected, it'll remain selected after I hit <kbd>Esc</kbd>. But it's no big deal, I'm fine with this as is if you prefer it that way.

dagheyman:
> Oh right, it's a different action. I meant just removing the keyboard shortcut. In that case perhaps we should revert to not focusing the search field by default.

Let's leave it as it is for now. We'll keep the space-shortcut and keep search focused by default. The shortcut will not work when the search has focus (a space will be printed instead), but it will work after tabbing out or selecting a credential with the mouse. I think searching and filtering for credentials is a more common use case than generating a code without copying it, so it make sense to have that require less steps.

PR #228, opened 2018-01-03
Improve cleanup on removed device

dagheyman:
This makes the UI behave better when removing and re-inserting a device.

### DIFF
--- a/qml/main.qml
+++ b/qml/main.qml
@@ -577,10 +577,14 @@ ApplicationWindow {
         if (device.hasDevice && ccidModeMatch) {
             device.promptOrSkip(passwordPrompt)
         } else {
+            // Device is removed, cleanup.
             passwordPrompt.close()
             setPassword.close()
             addCredential.close()
             addCredentialSlot.close()
+            touchYourYubikey.close()
+            credentials = null
+            selectedKey = null
         }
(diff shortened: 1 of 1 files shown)


qml/main.qml
```diff
@@ -577,10 +577,14 @@ ApplicationWindow {
         if (device.hasDevice && ccidModeMatch) {
             device.promptOrSkip(passwordPrompt)
         } else {
+            // Device is removed, cleanup.
             passwordPrompt.close()
             setPassword.close()
             addCredential.close()
             addCredentialSlot.close()
+            touchYourYubikey.close()
+            credentials = null
```
emlun: Isn't `credentials = null` already done here?
https://github.com/Yubico/yubioath-desktop/blob/773e12b6c41f4530c499e44a362dbd1865d307f6/qml/main.qml#L172

dagheyman: Yes, but it's to late, that's when the credentials have already been refreshed. That a device is not present is recognized earlier than that. So when switching devices the old credentials will be visible for half a second or so if we're not clearing the credentials instantly.

emlun: Ah, I see. Good catch, I hadn't noticed that.

PR #230, opened 2018-01-03
Handle undefined version string

dagheyman:


dagheyman:
Updated the commit.

Some context:
Without this, when starting the app in Slot Mode and then switching to CCID mode, the version variable is not assigned in time before the join, resulting in 
`qrc:///qml/utils.js:2: TypeError: Cannot call method 'join' of null` 
being printed to stdout.

PR #233, opened 2018-01-15
Hide global progress bar when no non-touch TOTP credentials is there

dagheyman:
When no non-touch TOTP credential is present, there is no need to show the global progress bar. It's distracting!

### DIFF
--- a/qml/YubiKey.qml
+++ b/qml/YubiKey.qml
@@ -208,6 +208,13 @@ Python {
         return false
     }
 
+    function hasAnyNonTouchTotpCredential() {
+        return util.find(entries, function (entry) {
+            return !entry.credential.touch
+                    && entry.credential.oath_type === 'TOTP'
+        }) || false
+    }
+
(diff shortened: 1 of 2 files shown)


qml/YubiKey.qml
```diff
@@ -208,6 +208,18 @@ Python {
...
+            for (var i = 0; i < entries.length; i++) {
+                if (!entries[i].credential.touch
+                        && entries[i].credential.oath_type === 'TOTP') {
+                    return true
+                }
+            }
+        }
+        return false
+    }
+
```
emlun: I added this search logic in `qml/Util.qml:find(items, predicate)`, might want to reuse that. Probably also merge `Util.qml` into `utils.js` (which I didn't notice at the time), but I can do that in a separate PR.

dagheyman: Cool, i'll look into it.

dagheyman:
Commit updated.

PR #234, opened 2018-01-15
Merge Util.qml into utils.js

emlun:


qml/utils.js
```diff
@@ -8,3 +8,41 @@ function versionGE(version, major, minor, micro) {
   }
   return false
 }
```
emlun: @dagheyman Should I change to 4-space indentation here too?

dagheyman: I just ran autoformat on it, it did some things.

PR #240, opened 2018-01-23
Remove some duplicated and dead code

dagheyman:


py/yubikey.py
```diff
@@ -152,19 +152,17 @@ def calculate(self, credential, timestamp):
...
-                slot, challenge=timestamp, totp=True, digits=int(digits),
-                wait_for_touch=True)
+            code = self._read_slot_code(
+                slot, digits, timestamp, wait_for_touch=True)
+            return pair_to_dict(Credential(self._slot_name(slot),
+                                           OATH_TYPE.TOTP, True), code)
+        except YkpersError as e:
+            if e.errno == 4:
+                logger.debug(
+                    'Time out error, user probably did not touch the device.')
```
emlun: Looks like we need an `else: logger.error(...)` here :)

dagheyman: Good catch, I thought the general Exception below would catch it, but I realise now that it doesn't.

emlun: I thought for a bit that maybe you could just re-throw the same exception, but I suppose that would just throw upwards and past the following `except` block instead.

Issue #224, opened 2017-12-20
"No YubiKey detected." on built-from-source Linux 4.3.0


replaid:
- Yubico Authenticator version: 4.3.0, built from source
- Operating system and version: PureOS 8.0
- YubiKey model and version: YubiKey 4
- Bug description summary: "No YubiKey detected." on built-from-source Linux 4.3.0

Steps to reproduce
---

1. Open `yubioath-desktop`, either from the command line or through the application launcher.
2. See message "No YubiKey detected."
3. Insert YubiKey into a USB port.

Expected result
---

Some behavior involving the "No YubiKey detected." message disappearing.

Actual results
---

The message remains with no changes in the app whatsoever.

Other info
---

This YubiKey works with the corresponding app on my Mac, and is recognized by the personalization tool on this Linux machine. See #223 for more back story in case helpful.

replaid:
The above happened with the 4.3.0 source download. I just reproduced with 96d11c2f9004640c3781477705b1fa07e85e1e8a.

replaid:
Thanks for your quick reply. I get the exact same behavior and zero terminal output when I run that command.

emlun:
That's strange. Do you have the YubiKey Manager python library (`ykman`) installed?

replaid:
Doesn't look like it.

```
$ python -c "import ykman"
Traceback (most recent call last):
  File "<string>", line 1, in <module>
ImportError: No module named ykman
$ echo $?
1
```

emlun:
Sorry, I forgot to mention it's a Python 3 library, so please check that too. If you don't have it, then that's at least one problem. It's still strange that you don't get any terminal output at all from `yubioath-desktop`, but I have no idea what might be causing that.

YubiKey Manager is included in the `vendor/yubikey-manager` directory in the source tarball, and those sources were enough for me to build the `yubikey-manager` and `python3-yubikey-manager` packages on Debian Buster:

```
$ cd vendor/yubikey-manager
$ debuild -us -uc
$ sudo dpkg -i ../python3-yubikey-manager_0.5.0_all.deb
```

After installing `python3-yubikey-manager`, the `yubioath-desktop` binary built by `qmake && make` successfully detects my YubiKey in Debian Buster.

Does any of that help?

replaid:
Thank you for that information, which supplied the clues I needed. I now have a working Yubico Authenticator app after spending several hours on this project.

[...]

PR #242, opened 2018-01-24
Improve build docs in general and for Debian-based dists in particular

### LINKED ISSUE #223: Documenting build on a Debian-based system
Linux user experience report.

Experience: I run a Debian-based distro. I got a GPG error when adding the Ubuntu PPA repo, so backed out of that approach since I wasn't sure it was the right one for a non-Ubuntu Debian-based distro. 

Feedback: if non-Ubuntu distros can use the Ubuntu PPA repo, please say so. (And let me know if this is the case so I can provide the error message I get if needed.)

Experience: So I turned to the build instructions to build 4.3.0. The needed dependencies are listed in prose, which provides the needed information to solve the problem, but I had to do numerous web searches etc. to figure out what specific packages to install.

Unfortunately I don't have my shell history, but I believe this is the list of packages I needed:

* `qt5-make`
* `qt5-default`
* `qtdeclarative5-dev`
* `qml-module-qtquick-controls`
* `qml-module-qtquick-dialogs`
* the line `deb http://deb.debian.org/debian experimental main` in `/etc/apt/sources.list`
* `qml-module-io-thp-pyotherside`

Feedback: please mention these specific dependencies in your build instructions.

Experience: I had to scrounge around the Internet to figure out how to get an application icon.

Feedback: Please mention in the build instructions that I can copy the icon to `~/.icons/` (or other location you recommend), the `.desktop` file to `~/.local/share/applications/`, and a reminder that `yubioath-desktop` needs to be in my `$PATH` for the icon to appear.

Experience: This was a lot of steps.

Feedback: please provide a Debian package.

And then after all this, it didn't see my YubiKey, even though YubiKey Personalization Tool (which I was able to install through my OS's GUI software install app) does. Filing a separate bug report #224 for that.

emlun:
Documentation improvements based on the resolution of #224 and the feedback in #223.

This fixes #223.

### DIFF
--- a/doc/Building_from_Source.adoc
+++ b/doc/Building_from_Source.adoc
@@ -12,6 +12,11 @@ To build from source, you need the following tools and dependencies:
 - Qt 5 development tools, including `qmake`
 - Qt QML development tools
 
+You can use the `Build-Depends` field in the
+https://github.com/Yubico/yubioath-desktop/blob/master/debian/control[Ubuntu
+package control file] as a guide to identifying which of your distribution's
+packages you need.
+
 With these installed, the build procedure is:
 
     $ git clone --recurse-submodules https://github.com/Yubico/yubioath-desktop.git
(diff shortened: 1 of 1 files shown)


emlun:
I agree about the mixed feelings. On the one hand they are indeed external dependencies in a conceptual sense - but on the other hand we do depend on specific git versions of them, most of them are Yubico projects (as opposed to widespread projects like Qt), the build definition requires them to be extracted to particular locations; so it makes some sense to include them in the tarball. But then again, there still remain other dependencies that are not included...

PR #249, opened 2018-01-30
Fix error in QR parsing

dagheyman:


qml/AddCredential.qml
```diff
@@ -206,36 +206,33 @@ DefaultDialog {
...
+        if (type === null || type === 'TOTP') {
             return 0
         }
         if (type === 'HOTP') {
             return 1
         }
-        return false
     }
 
     function getDigitsIndex(digits) {
```
emlun: I feel like we should probably `parseInt` here instead...

dagheyman: Not following, parse what int?

emlun: I mean in `getDigitsIndex`, below this message. :) Instead of matching explicitly on string values, I mean.

dagheyman: Sorry, still not following :) After the change we are comparing ints (digits) with ints (index) to preselect the correct option in the dropdown.

emlun: Um, never mind me. Apparently I went colorblind for a bit there. :D

PR #250, opened 2018-01-30
Fix expiration on single touch credential

dagheyman:
With a single touch credential the code would not gray out and the timer would not go away.

(This change seems to fix the issue, but there might be cleaner ways to solve this..)


dagheyman:
Closing, let's work with #251

PR #251, opened 2018-01-30
Fix expiration on single touch credential, alternative solution

emlun:
This feels like a cleaner solution than #250.

Basically it seems like the QML engine doesn't detect that the value of the expression `appWindow.isExpired(modelData)` could have changed, because it's not a functionally pure expression. If we instead base the property bindings on a property containing the current time, and imperatively update the time property instead, that should solve the issue.

We might want to consider extracting a "current time" property like this up to the top level and base all time based computations off that instead of `Date.now()`.

### DIFF
--- a/qml/CredentialItem.qml
+++ b/qml/CredentialItem.qml
@@ -17,8 +17,8 @@ Rectangle {
     readonly property bool hasCustomTimeBar: (code
                                               && ((credential.period
                                                    && credential.period !== 30)
-                                                  || (credential.touch
-                                                      && !isExpired)))
+                                                  || credential.touch))
+    readonly property bool showCustomTimeBar: hasCustomTimeBar && !isExpired
     readonly property bool isExpired: timeLeft <= 0
     readonly property color textColor: (isSelected ? palette.highlightedText : palette.windowText)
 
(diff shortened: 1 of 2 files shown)


dagheyman:
Agree that it's cleaner. 

I think I found a small issue though: When I have a list with 2 expired touch credentials, and re-generate one of them, the custom timer shows up for a small second on the other one as well. Then it goes away.

This doesn't happen in #250

emlun:
Hm, strange, I don't see that behaviour on my end. Maybe it's because `isExpired: timeLeft < 0` and not `isExpired: timeLeft <= 0`?

I also noticed a bug in that the timer bar would get stuck at 100% if the timer starts counting down from greater than the credential period (when we generate future codes instead of ones that soon expire).

qml/CredentialItem.qml
```diff
@@ -19,7 +19,7 @@ Rectangle {
                                                    && credential.period !== 30)
                                                   || (credential.touch
                                                       && !isExpired)))
-    readonly property bool isExpired: timeLeft < 0
+    readonly property bool isExpired: timeLeft <= 0
```
dagheyman: I think this change should be reverted to avoid the code starting out in a greyed out state.

emlun: Sure 👍

PR #252, opened 2018-01-31
Improve logging configuration

emlun:
Builds on https://github.com/Yubico/yubikey-manager/pull/46 .

- Fix issues with WARN and higher log levels being printed when `--log-level` CLI option is not given
- Ensure logging is initialized before loading the `yubikey` module


main.cpp
```diff
@@ -74,7 +74,13 @@ int main(int argc, char *argv[])
     QObject *root = engine.rootObjects().first();
 
     if (argc > 2 && strcmp(argv[1], "--log-level") == 0) {
-        QMetaObject::invokeMethod(root, "enableLogging", Q_ARG(QVariant, argv[2]));
+        if (argc > 4 && strcmp(argv[3], "--log-file") == 0) {
```
dagheyman: I guess this means that the arguments have to come in the correct order? I'm fine with this for a first version, but we might want to improve on that in the future.

emlun: Yes, this bugs me a bit too, but as long as these two are the only options I feel like the bother is less than the effort of setting up proper CLI parsing. We'll probably need to do that if we add another option, though.

PR #255, opened 2018-02-05
Fix packaging of source release tarball

### LINKED ISSUE #254: 4.3.2 release is NOT gz archive
it is just a tar file:

```
bash$ wget https://github.com/Yubico/yubioath-desktop/releases/download/yubioath-desktop-4.3.2/yubioath-desktop-4.3.2.tar.gz
bash$ file yubioath-desktop-4.3.2.tar.gz 
yubioath-desktop-4.3.2.tar.gz: POSIX tar archive (GNU)
```

emlun:
This fixes #254.

### DIFF
--- a/dev-util/package-sources.sh
+++ b/dev-util/package-sources.sh
@@ -70,7 +70,7 @@ for submodule_line in $(git submodule status --recursive); do
 done
 
 # Bundle all sources together
-tar cf "${output_archive_file_name}" -C "${OUTPUT_DIR}" .
+tar cf "${output_archive_file_name}" -C "${OUTPUT_DIR}" -a .
 
 echo "Wrote ${output_archive_file_name}"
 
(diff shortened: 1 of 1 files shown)


dagheyman:
This script is only run manually right? What about dropping the whole thing and just use `git archive --format=tar.gz` ? I don't think we should bundle submodules anyway.

emlun:
Yeah, if we drop bundling submodules we can do that. What do you think @dainnilsson?

dainnilsson:
The source tarball should be usable to build the binary artifacts, yes? Since the dependencies in the vendor directory are specific git commits (as opposed to properly versioned packages that you can point to), I would argue that they belong in the source release tarball.

dagheyman:
My opinion is that every released version (which would have a source package) should only depend on properly released and versioned packages. I think this would make things easier for 3rd party packagers as well.

dagheyman:
> By vendoring source code we make it part of our source release (and thus include it).

Thats fair. Maybe we should move away from the vendoring as well though. But that's outside the scope of this PR, so I'll approve it.

PR #258, opened 2018-02-07
Simplify Python module loading and initialization

emlun:
This solves some issues with doing the `site.addSiteDir` thing multiple times. `logging_setup.py` and `json_util.py` are now merged into `yubikey.py`, and the logging initialization race condition is instead resolved using a pair of initialization functions in the `yubikey` module.

### DIFF
--- a/py/yubikey.py
+++ b/py/yubikey.py
@@ -3,6 +3,8 @@
 
 import logging
 import types
+import ykman.logging_setup
+
 from base64 import b32encode, b64decode
 from binascii import a2b_hex, b2a_hex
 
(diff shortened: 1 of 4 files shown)


py/yubikey.py
```diff
@@ -322,4 +330,16 @@ def get_line(self, line_number):
             self.width * line_number:self.width * (line_number + 1)]
 
 
-controller = Controller()
+controller = None
+
+
+def initWithLogging(log_level, log_file=None):
```
dainnilsson: why_the_camel_case?

emlun: ConfusionBy_switching_backAndForth_between_pythonAndJavascriptProjects :)

emlun:
@dainnilsson we ok?

PR #263, opened 2018-02-13
Add shim for String.prototype.startsWith

dagheyman:
Without this, on Ubuntu Xenial we get `qrc:///qml/YubiKey.qml:65: TypeError: Property 'startsWith' of object yubikey.controller.count_devices is not a function`.

### DIFF
--- a/qml/utils.js
+++ b/qml/utils.js
@@ -12,6 +12,12 @@ function versionGE(version, major, minor, micro) {
     return false
 }
 
+// Shim for String.prototype.startsWith(), added in Qt 5.8
+// Reference: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/startsWith
+function startsWith(string, search, pos) {
+    return string.substr(!pos || pos < 0 ? 0 : +pos, search.length) === search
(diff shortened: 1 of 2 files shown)


qml/utils.js
```diff
@@ -12,6 +12,12 @@ function versionGE(version, major, minor, micro) {
     return false
 }
 
+// Shim for String.prototype.startsWith(), added in Qt 5.8
+// Reference: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/startsWith
+function startsWith(string, search, pos) {
+    return string.substr(!pos || pos < 0 ? 0 : +pos, search.length) === search
```
emlun: How about `Math.max(0, pos || 0)` for the first argument?

emlun: On second thought, when would you want to call this with `pos != 0`?

dagheyman: Support for position is in actual ES6 API, I'm fine with dropping it in our version if you prefer.

emlun: Ah, ok. I don't mind keeping it in.

PR #265, opened 2018-02-14
Make Windows version tweaking work with X.Y.Z-C-gCOMMIT[-dirty] version strings

emlun:
Previously, the code assumed that after removing a `-dirty` suffix there will be no more than 1 `-` in the remaining string. This is not true for, for example, `4.3.2-68-gCOMMIT-dirty`. This string matched both of the substitution rules, resulting in a Windows version of `4.3.2.0.1`
instead of the intended `4.3.2.0`.


emlun:
Um, apparently this isn't compatible with the current `-rc-`-something version. xD I'll fix that too.

PR #266, opened 2018-02-14
Detect and report unusable device to user

### LINKED ISSUE #264: Detect U2F key and report incompatibility
- Yubico Authenticator version: any
- Operating system and version: any
- YubiKey model and version: FIDO U2F Security Key
- Bug description summary: Yubico Authenticator attempts to connect to a FIDO U2F Security Key, and appears to hang because it can't open a connection.

Steps to reproduce
---

Start Yubico Authenticator and plug in a FIDO U2F Security Key

Expected result
---

The application reports that the FIDO U2F Security Key is incompatible with the app.

Actual results
---

The application hangs on "Connecting to YubiKey...".

emlun:
This fixes #264.

### DIFF
--- a/qml/utils.js
+++ b/qml/utils.js
@@ -19,6 +19,22 @@ function startsWith(string, search, pos) {
 }
 
 /*
+ * @param items: array
+ * @param sought: any
+ * @return True if and only if any `x === sought` for any `x` in `items`.
+ */
+function contains(items, sought) {
+    if (items) {
+        for (var i = 0; i < items.length; ++i) {
+            if (items[i] === sought) {
+                return true
+            }
+        }
+    }
+    return false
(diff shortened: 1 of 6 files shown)


qml/utils.js
```diff
@@ -18,6 +18,22 @@ function startsWith(string, search, pos) {
...
+    if (items) {
+        for (var i = 0; i < items.length; ++i) {
+            if (items[i] === sought) {
+                return true
+            }
+        }
+    }
+    return false
+}
+
```
emlun: Apparently I refactored the call to this out, so this is actually unused...

emlun: Apparently I refactored the call to this out, so this is actually unused...

PR #272, opened 2018-02-22
Replace QtSingleApplication with a QLockFile

dagheyman:
What:
- This PR drops the dependency on QtSingleApplication and instead uses a QLockFile to ensure that only one instance is running per user, as suggested [here](http://blog.aeguana.com/2015/10/15/how-to-run-a-single-app-instance-in-qt/). Showing a "App is already running" message seems to be a [common pattern](https://www.google.se/search?q=is+already+running&source=lnms&tbm=isch&sa=X&ved=0ahUKEwjI8930j7nZAhVEK1AKHZFHCEQQ_AUICigB&biw=1280&bih=704).

Why:
- Mostly to simplify the build process and allow us to drop the submodule. But the new behaviour should also make it a be bit more clear for the user what is going on. The previous behaviour on Windows was that the application started blinking in the start menu, but didn't actually showed itself, when the user tried to launch a new instance.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,3 +1,0 @@
-[submodule "vendor/qt-solutions"]
-	path = vendor/qt-solutions
-	url = https://github.com/qtproject/qt-solutions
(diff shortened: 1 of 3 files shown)


elboletaire:
You say that...

> The previous behaviour on Windows was that the application started blinking in the start menu, but didn't actually showed itself, when the user tried to launch a new instance.

But for me under windows was working as expected. Opening a new instance simply showed up the application, instead of blinking like you say :confused:

PR #278, opened 2018-03-09
Sort qrc input file list

bmwiedemann:
so that yubioath-desktop packages build in a reproducible way
in spite of indeterministic filesystem readdir order

See https://reproducible-builds.org/ for why this is good.

Note: With the current pretty flat resource layout,
dirs has at most 1 entry so would not need to be sorted,
but OTOH sorting this will be very fast
and it avoids future problems of this kind.


dagheyman:
Thanks, merged!

PR #294, opened 2018-05-28
Allow all three text fields in Add Credential dialog to submit the form

emlun:


dagheyman:
Can we split this up? I'm fine with the submit on enter-feature, but have some issues with the inline error message (hardcoded color, new UI element introduced, layout issues).

emlun:
Sure. I was thinking the latter was kind of necessary for the former to make any sense, but I guess without it it will just echo the button's behaviour.

emlun:
Replaced by #295.

@dagheyman Do you feel like there's anything to be salvaged from the inline error message thing, or do we just drop it?

dagheyman:
I suggest we drop it for now.

emlun:
:+1:

PR #314, opened 2018-11-06
slot mode: improve error handling

dagheyman:


qml/AddCredentialSlot.qml
```diff
@@ -146,15 +153,18 @@ DefaultDialog {
...
-                                         paddingError.open()
-                                     }
-                                     if (error === 'Over 20 bytes') {
-                                         tooLargeKeyError.open()
-                                     }
-                                     if (error) {
-                                         console.log(error)
+                                 touch.checked, function (resp) {
+                                     if (!resp.success) {
+                                         if (resp.error === 'Incorrect padding') {
```
emlun: I think it would probably be better to push these magic strings down into the Python layer and return more stable error identifiers from there. Matching like this seems brittle...

Ideal would of course be to get differentiable errors from the library, but that would take some more work.

dagheyman: I don't disagree with you, however I'm trying to follow the style we have in https://github.com/Yubico/yubikey-manager-qt/blob/master/ykman-gui/py/yubikey.py for now. Pushed some of the handling down, to allow shorter strings in QML layer.

emlun: Yeah, I've been struggling a bit with that too... :)

PR #337, opened 2019-03-18
fixed the search bar

braathen:


dagheyman:
Unfortunately this doesn't build with Qt 5.9 (the lowest version we currently support). This is because of the `placeholderTextColor : color` property, which was [introduced in Qt 5.12](https://doc.qt.io/qt-5/qml-qtquick-controls2-textfield.html) I'd suggest dropping that property for now, and then see if we can find a workaround or if we could raise the lowest supported version to 5.12 (it has other benefits as well).

Looks good otherwise!

PR #341, opened 2019-03-21
highlighting selected and key navigation

braathen:


qml/CredentialCard.qml
```diff
@@ -29,7 +29,20 @@ Pane {
...
+    }
+
+    MouseArea {
+        anchors.fill: parent
+        onClicked: credentialCard.GridView.isCurrentItem ? credentialCard.GridView.view.currentIndex = -1 : credentialCard.GridView.view.currentIndex = index
+    }
+
+    function itemColor() {
+        if (credentialCard.GridView.isCurrentItem) {
+            return app.isDark() ? "#444444" : "#eeeeee"
```
dagheyman: Maybe give these colors names as well, similar to the others?

braathen: agreed, fixing it

dagheyman:
Merging with a small fix. There seems to be an issue with clicking at the top or bottom edge not selecting the issue, but let's deal with that later.

PR #344, opened 2019-03-25
fixed selection close to edges

braathen:


qml/SettingsView.qml
```diff
@@ -13,8 +13,21 @@ Pane {
             text: "Settings"
         }
         Button {
-            text: "toggle theme"
+            id: button
+            flat: true
+            text: "<font color='#fefefe'>Toggle theme</font>"
```
dagheyman: Can't you use Material.foreground here ?

PR #386, opened 2019-05-28
changed and cleaned up some colors, fixed dialog and prompts

braathen:


qml/SettingsView.qml
```diff
@@ -94,26 +94,21 @@ ScrollView {
...
-                        console.log(resp.error_id)
-                    }
-                    clearPasswordFields()
-                })
-            } else {
-                navigator.snackBarError(getErrorMessage(resp.error_id))
-                console.log(resp.error_id)
-            }
-        })
+        if (resp.success) {
```
dagheyman: This change doesn't look right, `resp` is returned from `yubikey.py` , but it is being checked directly

PR #388, opened 2019-05-31
fixed flickering of no creds screen and added some shortcuts

braathen:


qml/main.qml
```diff
@@ -112,6 +112,35 @@ ApplicationWindow {
...
 
+    Shortcut {
+        sequence: StandardKey.Copy
+        enabled: currentCredentialCard !== undefined && currentCredentialCard
+        onActivated: app.currentCredentialCard.calculateCard(true)
+    }
+
+    Shortcut {
+        sequence: StandardKey.Delete
+        enabled: currentCredentialCard !== undefined && currentCredentialCard
```
dagheyman: You can shorten this to `!!currentCredentialCard` to check for both null and undefined.

PR #399, opened 2019-06-24
fullscreen mode and dropped hashing

braathen:


qml/main.qml
```diff
@@ -207,6 +211,25 @@ ApplicationWindow {
...
+    }
+
+    Shortcut {
+        sequence: StandardKey.FullScreen
+        onActivated: visibility = visibility
+                     === Window.FullScreen ? Window.Windowed : Window.FullScreen
+        context: Qt.ApplicationShortcut
+    }
+
+    Shortcut {
```
dagheyman: I don't think we need the hide toolbar feature

PR #431, opened 2019-09-09
New UI based on Quick Controls 2

dagheyman:
This is the work that will eventually become the 5.0.0 release.

- New design based on Material Style
- Light mode and Dark mode
- Add support for custom CCID readers
- New tray functionality
- Support for favorite credentials
- Support for switching between devices


py/yubikey.py
```diff
@@ -77,275 +82,437 @@ def credential_data_to_dict(credentialData):
...
+        valid_from = timestamp - (timestamp % 30)
+        valid_to = valid_from + 30
+        with self._open_otp() as otp_controller:
+            code = otp_controller.calculate(
+                slot, challenge=timestamp, totp=True,
+                digits=int(digits), wait_for_touch=True)
+            return success({
+                'credential': credential,
+                'code': code_to_dict(Code(code, valid_from, valid_to))
+            })
```
emlun: Would it make sense to use `_otp_get_code_or_touch()` here as well (even if we're assuming it should always return `touch=False`)?

PR #442, opened 2019-09-24
actions: better names for artifacts

dagheyman:


.github/workflows/appimage.yml
```diff
@@ -12,7 +12,7 @@ jobs:
 
       - name: Build docker image that produces the AppImage
         run: |
-          export REF=$(echo ${GITHUB_REF} | tr / -)
+          export REF=$(echo ${GITHUB_REF} | cut -d '/' -f 3)
```
emlun: You can also do it like this:

```
export REF=${GITHUB_REF#refs/*/}
```
See "remove matching prefix pattern" in the bash manual. :slightly_smiling_face:

PR #444, opened 2019-09-24
Use relative font sizes where possible

bendem:
I am no QML master, but the new design was unreadable to me. I had to remove the hardcoded font size to be able to use the app again.
This is not polished, if there is interest for it, I'll look into fetching the system's font size by default and adding a configuration option in the settings.
Also, I couldn't get the search input to play nicely so its placeholder is not horizontally aligned. If someone with experience knows what's up, feel welcome to point it out.

### DIFF
--- a/qml/AboutPopup.qml
+++ b/qml/AboutPopup.qml
@@ -69,7 +69,6 @@ Dialog {
 
             Label {
                 text: !!yubiKey.currentDevice ? yubiKey.currentDevice.name : ""
-                font.pixelSize: 13
                 font.weight: Font.Medium
                 Layout.alignment: Qt.AlignHCenter | Qt.AlignVCenter
                 width: parent.width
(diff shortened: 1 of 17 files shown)


bendem:
Screenshot before/after:
Screenshot 2019-09-24 at 16 05 18
Screenshot 2019-09-24 at 16 10 41

Screenshot 2019-09-24 at 16 12 25
Screenshot 2019-09-24 at 16 12 39

dagheyman:
Thanks for the contribution! There is probably room for both increasing the default size a bit, and respecting the system font size. I'd like to avoid an extra setting if possible. It's a good direction, we just need to make sure it plays well across different OSes and resolutions.

dagheyman:
Related: https://github.com/Yubico/yubioath-desktop/issues/307#issuecomment-534863764

bendem:
I've removed the pixelSize definition on the root component. That should use the system's font.
I can't test it on linux right now and I don't have Windows to test it on so would love it if someone could test it there.

dagheyman:
Screenshots from testing on Ubuntu 19.04. Will be able to test on Windows a bit tomorrow probably.

- The main view looks pretty good I think
- In settings, the describing text is no longer smaller than the heading, which it should be.
- In add credential, the first heading is cropped

bendem:
Made the header wrap and increased the font size of titles in expandable sections

dagheyman:
Hi, thanks for the contribution again. We experimented a bit with both using `font.pointSize` and relative sizes, but the result differs to much on different platforms, and makes the design harder to control. We will not merge this, but we have done some work to increase the font sizes, and bundle the font on all platforms for consistency. That will be included in the next version of the app.

PR #445, opened 2019-09-25
CI: drop travis (replaced with actions)

dagheyman:
- macOS build is moved to actions
- source package is moved to actions
- appimage build is moved to actions
- debian package build is dropped (was failing silently anyway, easy to re-introduce if needed)


dagheyman:
It does, and we do, but I'm thinking we should add a generic `build` workflow first, since there is no badge that represents all workflows. One badge per workflow is a bit much.

emlun:
Yeah, I was thinking the same - none of the three is an obvious choice for representing the build status as a whole.

PR #460, opened 2019-10-04
Actions: windows version done

fdennis:


.github/workflows/windows.yml
```diff
@@ -0,0 +1,110 @@
+name: windows
+
+on: [push]
```
dagheyman: add pull_request so it runs on PRs

.github/workflows/windows.yml
```diff
@@ -0,0 +1,110 @@
...
+          Copy-Item .\pymodules .\release -recurse
+
+          Get-ChildItem -File -Include *.pyc -Recurse | Remove-Item -Force
+          Get-ChildItem -Include __pycache__ -Recurse | Remove-Item -Force
+          Get-ChildItem -Include *.cpp -Recurse | Remove-Item -Force
+          Get-ChildItem -Include *.obj -Recurse | Remove-Item -Force
+          
+          7z a yubioath-desktop-test-win.zip release
+          mkdir deploy
+          cp yubioath-desktop-test-win.zip deploy
```
dagheyman: name the artifact yubioath-desktop-$BRANCH-win64.zip

dagheyman: (look at how the REF value is done in the other workflows)

dagheyman:
Great! Added two comments

.github/workflows/windows.yml
```diff
@@ -0,0 +1,110 @@
...
+        shell: powershell
+
+      - name: Upload artifact
+        uses: actions/upload-artifact@v1
+        with:
+          name: yubioath-desktop-app-bundle
```
dagheyman: also use another name for the artifact here, like `yubioath-desktop-win64`

PR #466, opened 2019-10-08
actions: use qt 5.13.1 on windows

dagheyman:


dagheyman:
Good catch. Feel free to clean them up in a separate PR :)

PR #480, opened 2019-10-22
remake of favorites and minor fixes

braathen:


qml/CredentialCard.qml
```diff
@@ -381,13 +371,65 @@ Pane {
         Accessible.name: (credential.issuer ? credential.issuer : credential.name)
         Accessible.description: getCodeLblValue()
 
+        ToolButton {
+            id: favoriteBtn
+            Layout.alignment: Qt.AlignRight | Qt.AlignTop
+            visible: favorite || credentialCard.hovered || credentialCard.GridView.isCurrentItem
+//            visible: true
```
dagheyman: Drop comment

qml/main.qml
```diff
@@ -223,6 +226,12 @@ ApplicationWindow {
         onActivated: app.currentCredentialCard.deleteCard()
     }
 
+    Shortcut {
+        sequence: "Ctrl+D"
```
dagheyman: Maybe add a small comment that this is Cmd + D on macOS

PR #494, opened 2019-11-27
opacity color method and prompt for copy

braathen:
almost entire color system re-done to a more Material Design inspired opacity based method, also added the final version of the code notification with prompt to create a copy when adding accounts.


fdennis:
I tested the latest commit. Looks good! Ready to merge, I think

PR #496, opened 2019-11-29
Allow spaces in manual secret key input field

alexwh:
Many sites put spaces in their secret keys for readability, and since we later strip spaces: 

https://github.com/Yubico/yubioath-desktop/blob/24ae87a286bc564a6f1afdfc5a696fa65a893c5c/py/yubikey.py#L339 

https://github.com/Yubico/yubikey-manager/blob/1f22620b623c6b345dd9f9193ec765a542dddc80/ykman/util.py#L458

...just accept them anyway.

### DIFF
--- a/qml/NewCredentialView.qml
+++ b/qml/NewCredentialView.qml
@@ -30,11 +30,13 @@ Flickable {
     contentHeight: content.implicitHeight + dynamicMargin
 
     function acceptableInput() {
+        // trim spaces to accurately count length, parse_b32_key later trims them
+        var secretKeyTrimmed = secretKeyLbl.text.replace(/ /g, "")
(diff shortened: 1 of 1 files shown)


alexwh:
Got it - I don't seem to have that button on my fork for some reason, but also wanted to avoid force pushes.

qml/NewCredentialView.qml
```diff
@@ -30,11 +30,13 @@ Flickable {
     contentHeight: content.implicitHeight + dynamicMargin
 
     function acceptableInput() {
+        // trim spaces to accurately count legnth, parse_b32_key later trims them
```
fdennis: Please fix the typo here also.
Typo: legnth -> length

PR #509, opened 2020-01-08
In preparation for next ykman cli release: Bumped python and minor changes to actions

fdennis:


dagheyman:
Looks good, is it on purpose to move the copying of the .DLLs to after `windeployqt` ?

fdennis:
> Looks good, is it on purpose to move the copying of the .DLLs to after `windeployqt` ?

Yes. There was a reason for doing so, but I'm pretty sure it doesn't apply anymore. Will change it back. Good catch!

.github/workflows/windows-x64.yml
```diff
@@ -105,7 +109,7 @@ jobs:
 
       - name: Create an unsigned .msi installer package
         run: |
-          $env:PATH += ";$env:WIX\bin"
+          $env:PATH +=";$env:WIX\bin"
```
emlun: While you're at it, can you restore this space too? :smile:

emlun:
> There was a reason for doing so, but I'm pretty sure it doesn't apply anymore.

Looks like the reason was that the `.\release\pymodules\ykman\native` directory doesn't exist before the windeployqt step?

emlun:
Shall we pull this in too? 51b9bd90196bff0ac9b08e10b424b8b94666a4b1

fdennis:
> Shall we pull this in too? [e9d0833](https://github.com/Yubico/yubioath-desktop/commit/e9d0833e8df3b16853b01364f58df1b043dce066)

Yes! Maybe make a PR to the dll-test-auth branch?

emlun:
I'll just push it in :)

.github/workflows/windows-x64.yml
fdennis: Restored all the spaces :)

PR #519, opened 2020-01-16
Rename icon to more safe name with domain name

ai:
Backport changes from [flatpak package](https://github.com/flathub/flathub/pull/1126/files#diff-510f50ae7bdd2b391150e0b4f74c9b1fR235) of your awesome tool. 

This PR will make Flatpak porting easier and use the latest Linux desktop app guidelines in icon naming (even outside of Flatpak world).

Most GNOME official apps use the same name format with the reverse domain as described in [app ID guidelines](https://docs.flatpak.org/en/latest/conventions.html#application-ids) and [FreeDesktop AppStream spec](https://www.freedesktop.org/software/appstream/docs/chap-Metadata.html#sect-Metadata-GenericComponent).


ai:
@dagheyman CI falls because of missing new Python 3.8 on some platform. Do I need to fix it in this PR?

dagheyman:
Thanks for the PR!
I'm not opposed the change as such, but we haven't historically used `YubiOath` in camel case much. Would `com.yubico.yubioath` work instead? It's what we currently use as the bundle id string on macOS.

>  CI falls because of missing new Python 3.8 on some platform. Do I need to fix it in this PR?

No, it's not related, we are looking in to that as well.

ai:
> Would com.yubico.yubioath work instead?

Fixed. Good idea, now it looks much cleaner.

dagheyman:
Great, thanks!

PR #520, opened 2020-01-17
Use SVG icon in Linux

ai:
Fix https://github.com/Yubico/yubioath-desktop/issues/518

@dagheyman thanks for very fast replies. I am not sure about `appDir`. Can you double-check it?


dagheyman:
The appDir is for the AppImage, and I just tested the AppImage built from your PR (found here https://github.com/Yubico/yubioath-desktop/pull/520/checks?check_run_id=394777294), seems to be fine. Merging, thanks!

PR #523, opened 2020-01-20
Mention pcscd dependency in Linux install instructions

emlun:
Looks like I forgot to make a PR of this... :smile:

Mirrors Yubico/yubikey-manager-qt#216


README
```diff
@@ -23,6 +23,12 @@ which can be done by using the https://developers.yubico.com/yubikey-manager-qt/
 
 Downloads for all supported operating systems are available https://developers.yubico.com/yubioath-desktop/Releases/[here].
 
+
+==== Linux
```
dagheyman: Can we add this as a single sentence under System Requirements instead?
`On Linux systems, make sure the pcscd service is installed and running`.

emlun: Sure!

PR #524, opened 2020-01-20
Appdata

ai:
Another backport from [Flatpak PR](https://github.com/flathub/flathub/pull/1126/) — AppStream metadata.

I also renamed `.desktop` file to reverse DNS naming (because of the same reasons behind [renaming the icon](https://github.com/Yubico/yubioath-desktop/pull/520), hope I didn’t forget to change path somewhere as I did it in the previous PR 😓).

**Why this metadata should be in upstream?**
* It is Linux FreeDesktop [standard](https://www.freedesktop.org/software/appstream/docs/chap-Quickstart.html)
* Many different platforms use it. For instance, Snap (Flatpak alternative from Canonical) [uses it too](https://snapcraft.io/docs/using-external-metadata).
* It was created as part of an agreement between major Linux vendors (i.e. Red Hat, Canonical, SUSE, Debian, Mandriva, etc.) to create an infrastructure for application installers.
* I think you may want to control brand-related things like screenshot or description.

**Is it correct?**
* I used [AppStream spec](https://www.freedesktop.org/software/appstream/docs/chap-Quickstart.html) and [Flatpak guidelines](https://github.com/flathub/flathub/wiki/AppData-Guidelines). The only part I missed was the “16:9 aspect ratio” for a screenshot. A lot of apps avoid this rule as well.
* I called `flatpak run org.freedesktop.appstream-glib validate` validator.

**Notes about the content of metadata:**
* I used `CC0-1.0` for `metadata_license` according to very free license recommendations to be able to use description/screenshot in different stores and cases.
* I used `Yubico` name in `developer_name`. Should we rename it to `Yubico developers` or to your name?
* I used my email in `update_contact` as a promise to help you with a file in the future. But we can replace it with another contact.
* I put only the latest `5.0.1` and `5.0.0` notes to `releases`. Technically, we can remove `releases`. But it will be so nice to see changes directly in Linux’s apps store.


resources/com.yubico.yubioath.appdata.xml
```diff
@@ -0,0 +1,66 @@
...
+  </description>
+
+  <screenshots>
+    ​<screenshot type="default">
+      <image type="source" width="850" height="393">https://raw.githubusercontent.com/Yubico/yubioath-desktop/master/screenshot.png</image>
+      <caption>Different modes and themes</caption>
+    </screenshot>
+  </screenshots>
+
+  <releases>
```
dagheyman: I would drop the releases section for now and keep this minimal, since we already maintain the NEWS file.

resources/com.yubico.yubioath.appdata.xml
```diff
@@ -0,0 +1,66 @@
...
+  <metadata_license>CC0-1.0</metadata_license>
+  <project_license>BSD-2-Clause</project_license>
+
+  <name>Yubico Authenticator</name>
+  <summary>Graphical interface for displaying OATH codes with a YubiKey</summary>
+  <url type="homepage">https://developers.yubico.com/yubioath-desktop/</url>
+  <content_rating type="oars-1.1"/>
+
+  <developer_name>Yubico</developer_name>
+  <update_contact>andrey@sitnik.ru</update_contact>
```
dagheyman: Put dag@yubico.com here

dagheyman:
Looks good! See comments

ai:
> Looks good! See comments for feedback.

Fixed 😺

resources/com.yubico.yubioath.appdata.xml
```diff
@@ -0,0 +1,36 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<component type="desktop-application">
+  <id>com.yubico.yubioath</id>
+  <launchable type="desktop-id">com.yubico.yubioath.desktop</launchable>
+
+  <metadata_license>CC0-1.0</metadata_license>
```
dagheyman: Need to double check the license choice a bit internally before merging, but is probably fine.

dagheyman:
Very fast! Added a couple more :)

resources/com.yubico.yubioath.appdata.xml
```diff
@@ -0,0 +1,36 @@
...
+  <project_license>BSD-2-Clause</project_license>
+
+  <name>Yubico Authenticator</name>
+  <summary>Graphical interface for displaying OATH codes with a YubiKey</summary>
+  <url type="homepage">https://developers.yubico.com/yubioath-desktop/</url>
+  <content_rating type="oars-1.1"/>
```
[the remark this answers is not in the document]
ai: It is part of Flatpak guideline https://github.com/flathub/flathub/wiki/AppData-Guidelines#oars-information.

It means that the app could be used for children or users who want to avoid NSFW content (like PEGI rating system).

resources/com.yubico.yubioath.appdata.xml
ai: Sure, no rush

resources/com.yubico.yubioath.appdata.xml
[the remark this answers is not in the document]
dagheyman: Thanks, it can stay in then.

resources/com.yubico.yubioath.appdata.xml
dagheyman: @dainnilsson any input here? https://www.freedesktop.org/software/appstream/docs/chap-Metadata.html here are the supported licenses, under `metadata_license`

dainnilsson: If I'm reading this correctly, this license _only_ affects the metadata file itself? If so, I'd go with MIT as it's the one I'm most familiar with and it quite similar to the BSD license we use for the project.

dagheyman: Agreed, MIT is my suggestion as well. 

@ai Can you update the PR to use MIT for the metadata_license ?

ai: Fixed

resources/com.yubico.yubioath.appdata.xml
bilelmoussaoui: The releases are valuable info, it's used in Flathub/GNOME Software/KDE Discover/.. to show to the user which version is available/installed on their system. 

It's also a requirement for publishing an application on Flathub.

bilelmoussaoui: The NEWS file, is mostly read by: contributors, collaborators, developers or someone involved with the code source. The releases tags here targets a general audience (less technical) I would say

resources/com.yubico.yubioath.appdata.xml
```diff
@@ -0,0 +1,35 @@
...
+<component type="desktop-application">
+  <id>com.yubico.yubioath</id>
+  <launchable type="desktop-id">com.yubico.yubioath.desktop</launchable>
+
+  <metadata_license>MIT</metadata_license>
+  <project_license>BSD-2-Clause</project_license>
+
+  <name>Yubico Authenticator</name>
+  <summary>Graphical interface for displaying OATH codes with a YubiKey</summary>
+  <url type="homepage">https://developers.yubico.com/yubioath-desktop/</url>
```
bilelmoussaoui: It could use a bugtracker url

ai: It is not a field for issue tracker, it is homepage

resources/com.yubico.yubioath.appdata.xml
dagheyman: Didn't know that it was a requirement for Flathub. I'm happy to include these again as they were (sorry @ai ... ), and we'll look into later how to sync the content for them for upcoming releases.

resources/com.yubico.yubioath.appdata.xml
dagheyman: This repository, as well as the URL to open a support ticket ( https://support.yubico.com/support/tickets/new  ) is fine to include.

resources/com.yubico.yubioath.appdata.xml
bilelmoussaoui: It shouldn't be hard with https://blog.tenstral.net/2020/01/a-big-appstream-status-update.html. There's now a `news-to-metainfo` command shipped with appstream-util if you want to automate the task.

ai: I reverted `releases` removal and add fresh 5.0.2 release there 5883176

resources/com.yubico.yubioath.appdata.xml
ai: Replaced ccad977

dagheyman:
Thanks for all the work on this, merging!

PR #532, opened 2020-01-30
Use AppData in Linux builds

ai:
In #524 we added `appdata.xml` file to the repository. In this PR we are using it during install scripts for Linux.

* I added AppData to Debian script according to [Debian guideline](https://wiki.debian.org/AppStream/Guidelines#General)
* I didn’t check AppImageKit, but according to [it’s sources](https://github.com/AppImage/AppImageKit/blob/23267a1ed9276c4bdfb4d3a6137733a936f09490/src/appimagetool.c#L779) AppImageKit should load AppData from `appDir/`

@dagheyman We will need a release after merging this PR to start working for Flatpak.

### DIFF
--- a/debian/yubioath-desktop.install
+++ b/debian/yubioath-desktop.install
@@ -1,2 +1,3 @@
 resources/com.yubico.yubioath.desktop usr/share/applications
+resources/com.yubico.yubioath.appdata.xml usr/share/metainfo
 resources/icons/com.yubico.yubioath.svg usr/share/pixmaps
--- a/docker/xenial-appimage/Dockerfile
+++ b/docker/xenial-appimage/Dockerfile
@@ -46,6 +46,7 @@ RUN mkdir -p yubioath-desktop/appDir/usr \
     && qmake \
     && make \
     && cp resources/com.yubico.yubioath.desktop appDir/ \
+    && cp resources/com.yubico.yubioath.appdata.xml appDir/ \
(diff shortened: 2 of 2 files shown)


dagheyman:
Actually, we currently don't publish any debian packages ourselves (there is an effort to do it elsewhere though, see https://github.com/Yubico/yubioath-desktop/issues/328#issuecomment-568292005). So I was planning on removing the whole debian/ folder from the repository. Are those files used directly by Flatpak ?

The appimage part of the PR looks fine.

ai:
Do you build `deb` or `rpm`? I thought that AppImageKit is used for it.

dagheyman:
No, its only used for the .AppImage found on https://developers.yubico.com/yubioath-desktop/Releases/

ai:
It is strange that you build not so popular AppImage and not more popular snap or Flatpak 😅

OK, with this new information I need a day for investigating.

ai:
Seems like we do not need anything special for Flatpak. I will update `flathub/flathub` PR on weekend.

I think AppImage can still win from AppData: https://docs.appimage.org/packaging-guide/appstream.html

If we have `debian/` let’s update them as well. At least for theoretical consistency, rather than rational reasons.

But yeap, we do not need a release.

dagheyman:
Thanks for clarifying. Merging this, with the caveat that the whole debian directory will probably be removed soon.

PR #538, opened 2020-02-10
actions: build snap package

dagheyman:
Add initial snap build, based on qt 5.14 and custom python. I think this is ready for initial merging, even if there is some stuff left to do.

Testing:
```
$ sudo snap install yubioath-desktop_5.0.2_amd64.snap --dangerous
$ sudo snap connect yubioath-desktop:hardware-observe :hardware-observe
$ sudo snap connect yubioath-desktop:raw-usb :raw-usb
$ sudo snap connect yubioath-desktop:u2f-devices :u2f-devices
$ sudo snap connect yubioath-desktop:ykman-config :personal-files
$ sudo snap restart yubioath-desktop.pcscd
```

TODO:
- the bundled pcscd is now just installed and started, probably needs to be handled better. See https://forum.snapcraft.io/t/best-way-to-talk-to-smart-cards-pcscd/15318/
- all the manual connections above will hopefully go away after requesting auto connections. See https://snapcraft.io/docs/permission-requests

### DIFF
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -42,9 +42,6 @@ apps:
       - u2f-devices
       - hidraw
       - hardware-observe
-      - network
-      - network-bind
-      - network-observe
       - ykman-config
     extensions: [gnome-3-28]
 
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -110,7 +110,12 @@ parts:
     override-build: |
       eval "$(pyenv init -)"
       pyenv global 3.8.1
-      pip3 install git+git://github.com/dagheyman/pyscard.git#egg=pyscard
+      pip3 download --no-binary :all: pyscard
+      tar -xvf pyscard*
+      cd pyscard*
+      patch setup.py ../.github/workflows/snap-pyscard-patch.patch
+      pip3 install .
+      cd ..
       pip3 install -r requirements.txt
     override-stage: |
       snapcraftctl stage
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -1,0 +1,226 @@
+name: yubioath-desktop
+adopt-info: yubioath-desktop
+icon: resources/icons/com.yubico.yubioath.svg
+base: core18
+grade: devel
+confinement: strict
+
+plugs:
+  ykman-config:
+    interface: personal-files
+    write:
+      - $HOME/.ykman
+apps:
+  pcscd:
+    command: usr/sbin/pcscd --foreground --auto-exit
+    daemon: simple
+    common-id: com.yubico.pcscd
+    plugs:
+      - hardware-observe
+      - network
+      - network-bind
+      - raw-usb
+
+  yubioath-desktop:
+    adapter: full
+    command: yubioath-desktop
+    common-id: com.yubico.yubioath
+    environment:
+      LD_LIBRARY_PATH: $SNAP/opt/qt514/lib/x86_64-linux-gnu/:$SNAP/opt/qt514/lib/:$LD_LIBRARY_PATH:$SNAP/usr/lib/
+      QT_QPA_PLATFORM: xcb
+      QT_BASE_DIR: $SNAP/opt/qt514/
+      QT_PLUGIN_PATH: $SNAP/opt/qt514/plugins/
+      QML2_IMPORT_PATH: $SNAP/opt/qt514/qml/
+    plugs:
+      - desktop
+      - desktop-legacy
+      - wayland
+      - unity7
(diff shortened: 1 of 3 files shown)


snap/snapcraft.yaml
```diff
@@ -0,0 +1,231 @@
...
+      - $HOME/.ykman
+apps:
+  pcscd:
+    command: usr/sbin/pcscd --foreground --auto-exit
+    daemon: simple
+    common-id: com.yubico.pcscd
+    plugs:
+      - hardware-observe
+      - network
+      - network-bind
```
emlun: Offline conversation: double check that network plugs are needed

snap/snapcraft.yaml
```diff
@@ -0,0 +1,231 @@
...
+      - unity7
+      - opengl
+      - x11
+      - raw-usb
+      - u2f-devices
+      - hidraw
+      - hardware-observe
+      - network
+      - network-bind
+      - network-observe
```
emlun: Offline conversation: double check that network plugs are needed

snap/snapcraft.yaml
dagheyman: It seems to be needed, I get this when removing them from `pcscd`. It looks like I can drop them from the `yubioath-desktop` part though!
```
$ /snap/yubioath-desktop/current/usr/sbin/pcscd --foreground
00000000 pcscdaemon.c:698:main() cannot create /run/snap.yubioath-desktop/pcscd.pid: Permission denied
00000071 winscard_msg_srv.c:148:InitializeSocket() Unable to bind common socket: Permission denied
00000010 pcscdaemon.c:723:main() Error initializing pcscd.
```

snap/snapcraft.yaml
dagheyman: Dropped!

snap/snapcraft.yaml
emlun: :+1:

snap/snapcraft.yaml
```diff
@@ -0,0 +1,228 @@
...
+      - wayland
+      - unity7
+      - opengl
+      - x11
+      - raw-usb
+      - u2f-devices
+      - hidraw
+      - hardware-observe
+      - ykman-config
+    extensions: [gnome-3-28]
```
Maryse47: @dagheyman why qt app needs gnome extension?

dagheyman: It's a good question, I will investigate if it works with the `kde-neon` extension instead. Thanks!

Maryse47: According to https://snapcraft.io/docs/kde-neon-extension :

> This extension is currently only available for the KDE project itself. Snaps from non-KDE publishers will not get automatically connected to the KDE runtime snap required by this extension.

However if you build your own qt then maybe you don't need any extension?

dagheyman: Ah, that's probably why I didn't added it initially. Adding the gnome-extension definitely does something that is needed, but I need to investigate exactly what it does and if we could do the same without it.

Maryse47: Doc says it connects to [gtk-common-themes](https://snapcraft.io/gtk-common-themes) and [gnome-3-28-1804](https://snapcraft.io/gnome-3-28-1804) content snaps.  You may specify the former manually and the latter shouldn't be needed.

The kde-neon extension docs say:

> if your application needs access to the Qt5 and KDE Frameworks development tools, add kde-frameworks-5-core18-sdk to the build-snaps of the part that builds your application. Some applications also require the Qt runtime libraries to be present at build time. In that case, also add kde-frameworks-5-core18 to build-snaps.

So you may specify `gtk-common-themes` and `kde-frameworks-5-core18` instead of using `gnome-3-28-1804` extension.

Extensions also sets:

> 
>     desktop
>     desktop-legacy
>     wayland
>     x11
> 

but you set above explicitly in manifest anyway.

Maryse47: They also recommend:

> For a complete picture of what this extension does, add it to your app definition and run snapcraft expand-extensions.

dagheyman: Thanks for the pointers.

However I cant seem to get this to work without the gnome-extension currently, I'm getting all kinds of errors related to openGL and fonts without it. So I'm leaving it in for now.

Pull requests welcome though!

Maryse47: I'm not snap expert neither. I think if you post your manifest as _Call for testing_ you would get more eyeballs on this: https://forum.snapcraft.io/c/snap

dagheyman: Good idea! Might do that later actually, thanks.

PR #541, opened 2020-02-12
Make sure the application is always visible on a monitor

### LINKED ISSUE #241: Yubico Authenticator opens outside monitors and can not be accessed
Thank you for opening an issue! If this is a bug report, please include the
following information:

- Yubico Authenticator version: 4.3.2
- Operating system and version: Windows 10 pro
- YubiKey model and version: Neo
- Bug description summary: 

Yubico Authenticator opens outside monitors and can not be accessed anymore.

Steps to reproduce
---

I am using two 4k monitors with a resolution of 3840x2160 each, where 2 is my main monitor.

monitors

When opening Authenticator for the first time it is centered on my main monitor 2. I can move the Authenticator on monitor 2 and close it. Evereything works as expected and the settings are store in the registry as follows:

settings1

**Problem: When i move the Authenticator to my monitor 1 and i close the Authentcator on monitor 1 i am not able to see it anymore. I can see Authenticator starting and that it seems to be left outside of monitor 1.** Following settings are stored in the windows registry:

settings2

Expected result
---

I expect that Authenticator to open at monitor 1 (on same location where i closed it) next time i start the application.

Actual results
---

Authenticator opens outside left montior 1. I can not see it and can not use it. I need to manually delete the registry settings in my windows registry.

Other info
---

Great work and a great tool so far :-) Hopefully this description can help solving this issue. I can test your fix, if there will be one.

fdennis:
Needs testing on all OSes

Fixes #241

### DIFF
--- a/main.cpp
+++ b/main.cpp
@@ -1,4 +1,5 @@
 #include <QApplication>
+#include <QDesktopWidget>
 #include <QQmlApplicationEngine>
 #include <QQmlContext>
 #include <stdlib.h>
@@ -45,6 +46,21 @@ int main(int argc, char *argv[])
     application.setApplicationVersion(APP_VERSION);
     application.setOrganizationName("Yubico");
     application.setOrganizationDomain("com.yubico");
+
+    // Get x and y coordinates of all monitors
+    QVariantList monitorAreas;
+    for (QScreen* screen : QGuiApplication::screens())  {
+        QRect monitorArea = screen->geometry();
+
+        QVariantMap coordinates;
+
+        coordinates.insert("xMin", monitorArea.x());
+        coordinates.insert("xMax", monitorArea.x() + monitorArea.width());
+        coordinates.insert("yMin", monitorArea.y());
+        coordinates.insert("yMax", monitorArea.y() + monitorArea.height());
+
+        monitorAreas << coordinates;
+    }
 
     QQuickStyle::setStyle("Material");
 
@@ -97,6 +113,7 @@ int main(int argc, char *argv[])
     engine.rootContext()->setContextProperty("appVersion", APP_VERSION);
     engine.rootContext()->setContextProperty("ScreenShot", &screenshot);
     engine.rootContext()->setContextProperty("application", &application);
+    engine.rootContext()->setContextProperty("monitorAreas", monitorAreas);
     engine.load(QUrl(url_prefix + main_qml));
 
 
--- a/qml/main.qml
+++ b/qml/main.qml
@@ -139,6 +139,30 @@ ApplicationWindow {
                 && (settings.desktopAvailableHeight === Screen.desktopAvailableHeight)
         app.x = (savedScreenLayout) ? settings.x : Screen.width / 2 - app.width / 2
         app.y = (savedScreenLayout) ? settings.y : Screen.height / 2 - app.height / 2
+
+        function isWindowPositionInsideSomeMonitor() {
+            for (var i = 0; i < monitorAreas.length; i++)  {
+                var xMin = monitorAreas[i].xMin
+                var xMax = monitorAreas[i].xMax
+                var yMin = monitorAreas[i].yMin
+                var yMax = monitorAreas[i].yMax
+
+                if (app.x > xMin && app.x < xMax) {
+                    if (app.y > yMin && app.y < yMax) {
+                        return true
+                    }
+                }
+            }
+            return false
+        }
+
+        // If app.x and app.y are outside of the available screen geometry,
+        // put the app in the middle of the screen.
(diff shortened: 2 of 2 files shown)


main.cpp
```diff
@@ -46,6 +47,29 @@ int main(int argc, char *argv[])
     application.setOrganizationName("Yubico");
     application.setOrganizationDomain("com.yubico");
 
+    QDesktopWidget * desktop = QApplication::desktop();
+
+    // Get x and y coordinates of all monitors
+    QVariantList monitorList;
```
dagheyman: Maybe call this monitorAreas ?

main.cpp
```diff
@@ -46,6 +47,29 @@ int main(int argc, char *argv[])
...
+    for (qint32 i = 0, screens = desktop->screenCount(); i < screens; i++)  {
+        QRect monitor = desktop->screenGeometry(i);
```
dagheyman: This could then be called monitorArea

qml/main.qml
```diff
@@ -318,7 +336,9 @@ ApplicationWindow {
         interval: 1000
         repeat: true
         running: app.isInForeground || settings.closeToTray
-        onTriggered: yubiKey.poll()
+        onTriggered: {
```
dagheyman: Unrelated formatting change, remove

qml/main.qml
```diff
@@ -139,6 +139,24 @@ ApplicationWindow {
...
+            var miny = monitorList[i+1].first
+            var maxy = monitorList[i+1].second
+
+            if (app.x > minx && app.x < maxx) {
+                if (app.y > miny && app.y < maxy) {
+                    return
+                }
+            }
+        }
+
```
dagheyman: We should probably add a comment here, that this is the fallback when the saved position seems to be outside of the available screen geometry, and we then just put the app in the middle of the screen.

main.cpp
```diff
@@ -46,6 +47,29 @@ int main(int argc, char *argv[])
...
+
+        QVariantMap x;
+        QVariantMap y;
+
+        int monitor_x_min = monitor.x();
+        int monitor_x_max = monitor_x_min + monitor.width();
+        x.insert("first", monitor_x_min);
```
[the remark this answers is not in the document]
braathen: what if there's 3 (or more) monitors?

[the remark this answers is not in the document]
fdennis: > what if there's 3 (or more) monitors?

That does not matter. In this code we are storing the coordinates of every monitor. Every monitor has xmin, xmax, ymin, and ymax, since they are rectangles.

main.cpp
```diff
@@ -46,6 +47,29 @@ int main(int argc, char *argv[])
     application.setOrganizationName("Yubico");
     application.setOrganizationDomain("com.yubico");
 
+    QDesktopWidget * desktop = QApplication::desktop();
+
+    // Get x and y coordinates of all monitors
+    QVariantList monitorList;
+    for (qint32 i = 0, screens = desktop->screenCount(); i < screens; i++)  {
```
emlun: I'm getting a deprecation warning in Qt 5.14:

```
main.cpp:54:55: warning: ‘int QDesktopWidget::screenCount() const’ is deprecated: Use QGuiApplication::screens() [-Wdeprecated-declarations]
   54 |     for (qint32 i = 0, screens = desktop->screenCount(); i < screens; i++)  {
      |                                                       ^
```

Looks like the non-deprecated API is something like this:

```cpp
    for (QScreen* screen : QGuiApplication::screens())  {
        QRect geometry = screen->geometry();
        // ...
        int monitor_x_min = geometry.x();
        // ...
    }
```

main.cpp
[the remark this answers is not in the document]
emlun: With that, we can also put both `xMin`/`xMax` and `yMin`/`yMax` in the same `QVariantMap` instead of having to add two per monitor.

qml/main.qml
emlun: I agree, but I think even better would be to refactor the early return. For example by using a nested function:

```
        function windowPositionIsInsideSomeMonitor() {
            for (var i = 0; i < monitorList.length; i+=2)  {
                var minx = monitorList[i].first
                var maxx = monitorList[i].second
                var miny = monitorList[i+1].first
                var maxy = monitorList[i+1].second

                if (app.x > minx && app.x < maxx) {
                    if (app.y > miny && app.y < maxy) {
                        return
                    }
                }
            }
        }

        if (!windowPositionIsInsideSomeMonitor()) {
            app.x = Screen.width / 2 - app.width / 2
            app.y = Screen.height / 2 - app.height / 2
        }
```

main.cpp
fdennis: Good catch!

PR #544, opened 2020-02-19
mac: make app compatible with macOS app store

dagheyman:


.github/workflows/macOS.yml
```diff
@@ -41,44 +58,40 @@ jobs:
...
-          sudo find /usr/local/Cellar/openssl@1.1/ -name '*.dylib' -exec cp '{}' yubioath-desktop.app/Contents/Frameworks/ ';'
+          sudo find /opt/openssl/ -name '*.dylib' -exec cp '{}' yubioath-desktop.app/Contents/Frameworks/ ';'
 
       - name: Copy over Python.framework to app bundle
         run: |
           cp -a /Library/Frameworks/Python.framework yubioath-desktop.app/Contents/Frameworks/
           sudo find yubioath-desktop.app/Contents/Frameworks/Python.framework -name '*.pyc' -delete
           sudo find yubioath-desktop.app/Contents/Frameworks/Python.framework -name '__pycache__' -delete
 
-      - name: Move python dependencies to site-packages (required by codesign)
```
fdennis: Any reason why this was removed? Is it because we do not have to codesign when uploading to macOS app store?

dagheyman: We do it earlier in the build now, installing them in site-packages in the custom Python we compiled.

.github/workflows/macOS.yml
```diff
@@ -16,23 +17,39 @@ jobs:
...
         run: |
           brew update
-          brew upgrade
           brew bundle --file=./.github/workflows/Brewfile
           brew link qt -f
 
-      - name: Build custom Python as a framework
+      - name: Build OpenSSL
+        run: |
+          wget https://www.openssl.org/source/openssl-${OPENSSL_VER}.tar.gz
```
emlun: Should we have checksum verification for this, like for the windows builds in ykman-qt?

dagheyman: Yes, it's a good addition for all tarballs we download in this workflow, but let's do it in separate PRs.

PR #547, opened 2020-02-28
actions: improve dependency verification on macOS

dagheyman:
Use gpg signatures if available, otherwise sha256 checksums.

### DIFF
--- a/.github/workflows/macOS.yml
+++ b/.github/workflows/macOS.yml
@@ -23,12 +23,19 @@ jobs:
       - name: Install dependencies from homebrew
         run: |
           brew update
-          brew install swig qt zlib
+          brew install swig qt zlib gnupg curl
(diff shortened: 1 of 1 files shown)


.github/workflows/macOS.yml
```diff
@@ -60,6 +69,7 @@ jobs:
       - name: Build PyOtherSide QML plugin
         run: |
           wget https://github.com/thp/pyotherside/archive/${PYOTHERSIDE_VER}.tar.gz
+          echo "189cb0b973e40fcb6b95fd51b0bcd6cc8494b514d49ffe966ec488cf05bbf51e ${PYOTHERSIDE_VER}.tar.gz" | sha256sum -c -
```
emlun: Nice!

PR #552, opened 2020-03-05
Actions: Fixed aqtinstall

fdennis:
-Using an action that uses aqtinstall, instead of doing it all manually.

-Bumped aqtversion to 0.8a3, which has a fix for issues we encountered (QT could not be installed properly)


dagheyman:
Nice!

Might be interesting to use this action on macOS as well, to be able to lock the Qt version, but lets look into that separately.

PR #555, opened 2020-03-13
improve ccid error handling

dagheyman:
this change is meant to handle a couple of scenarios:

- No pcscd or smart card services is running at all. Catch `EstablishContextException` from pyscard if we want to handle it better later, requires some refactoring though. Currently this results in a generic error shown to the user, but the debug log should tell what is going on.

- Pcscd is running, but can't grab the device (LIBUSB_ERROR_BUSY) or list readers. This can happen in the snap version, since a competing pcscd might have grabbed it. Same behavior, just show an error to the user.

### DIFF
--- a/qml/YubiKey.qml
+++ b/qml/YubiKey.qml
@@ -239,7 +239,7 @@ Python {
                 if (!currentDevice || !availableDevices.some(dev => dev.serial === currentDevice.serial)) {
                     // new device is being loaded, clear any old device
                     clearCurrentDeviceAndEntries()
-
+                    navigator.goToLoading()
                     if (availableDevices.some(dev => dev.selectable)) {
                         // pick the first selectable device
                         currentDevice = resp.devices.find(dev => dev.selectable)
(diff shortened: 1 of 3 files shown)


qml/YubiKey.qml
```diff
@@ -309,6 +309,11 @@ Python {
                     currentDevice.hasPassword = true
                     currentDeviceValidated = false
                     navigator.goToEnterPasswordIfNotInSettings()
+                } else if (resp.error_id === 'open_device_failed') {
+                    clearCurrentDeviceAndEntries()
+                    yubiKey.availableDevices = []
+                    navigator.snackBarError("Failed connecting to YubiKey.")
```
fdennis: How about this instead? We have this format in a few places. I like it. open_device_failed will print ` Failed to connect to YubiKey`
```suggestion
                    navigator.snackBarError(navigator.getErrorMessage(
+                                            resp.error_id))
```

dagheyman: Good idea, changed

qml/YubiKey.qml
```diff
@@ -239,7 +239,7 @@ Python {
                 if (!currentDevice || !availableDevices.some(dev => dev.serial === currentDevice.serial)) {
                     // new device is being loaded, clear any old device
                     clearCurrentDeviceAndEntries()
-
+                    navigator.goToLoading()
```
fdennis: Is this the spinning circle? Nice touch!

dagheyman: It is, yes.

PR #557, opened 2020-03-16
Added yk5cnfc images

fdennis:


dagheyman:
LGTM! But did not test on macOS.

PR #567, opened 2020-04-07
highlight search for accounts and settings

braathen:


qml/StyledToolBar.qml
```diff
@@ -147,7 +147,9 @@ ToolBar {
                 selectedTextColor: defaultBackground
                 Layout.fillWidth: true
                 Layout.fillHeight: true
-                placeholderText: qsTr("Quick find")
+                placeholderText: !!(navigator.currentItem && navigator.currentItem.objectName !== 'settingsView')
+                                 ? qsTr("Quick find")
+                                 : qsTr("Quick find in Settings")
```
dagheyman, line 152: Sentence case? Maybe just "Search accounts" and "Search settings"?

braathen: agreed, fixed!

PR #589, opened 2020-04-29
1st stab at supporting all transports

dagheyman:
this is not complete, but good to merge to dev branch if it doesn't brake too much stuff.


dagheyman:
Thanks. This is still WIP I think, not mergeable right now, will create new PR later

PR #598, opened 2020-05-13
redesign of add flow and dialog UI

braathen:


dagheyman:
prints `qrc:/qml/StyledComboBox.qml:30: ReferenceError: noedit is not defined` when entering add credential mode

dagheyman:
Also looks like I get touch credentials with the checkbox unchecked

PR #602, opened 2020-05-18
Actions: DPI awareness to Win

fdennis:


.github/workflows/windows-x64.yml
```diff
@@ -119,6 +119,12 @@ jobs:
         run: |
           choco install visualstudio2019community --package-parameters "--add Microsoft.VisualStudio.Component.VC.Redist.MSM"
 
+      - name: DPI awareness
+        shell: powershell
+        run: |
+          $env:PATH += ";C:\Program Files (x86)\Windows Kits\10\bin\10.0.18362.0\x64"
+          mt.exe -manifest .\yubioath-desktop.exe.manifest -outputresource:.\release\yubioath-desktop.exe;#1
```
dagheyman, line 126: Should the `#1` in the end be there?

yubioath-desktop.exe.manifest
```diff
@@ -0,0 +1,51 @@
+<?xml version="1.0" encoding="UTF-8" standalone="yes"?>
```
dagheyman, line 1: Can we put this file in resources/win ?

.github/workflows/windows-x64.yml
fdennis: Yes, however, I don't think it's required.
https://docs.microsoft.com/en-us/windows/win32/sbscs/mt-exe

yubioath-desktop.exe.manifest
fdennis: Makes sense, will move

.github/workflows/windows-x64.yml
fdennis: Removed that part. Seems to work!

PR #607, opened 2020-05-22
Actions: bumped python to 3.8.3 and fix for mergemodules

fdennis:


dagheyman:
Builds doesn't seem to finish

fdennis:
> Builds doesn't seem to finish

Yeah. Probably nothing serious. I have restarted them

.github/workflows/windows-x64.yml
```diff
@@ -3,7 +3,7 @@ name: windows 64 bit
 on: [push, pull_request]
 
 env:
-  PYTHON_VER: '3.8.2'
+  PYTHON_VER: '3.x'
```
dagheyman, line 6: I guess this could brake when the action moves to 3.9 but the variable below is still 3.8 ?

fdennis: Yeah well I am just trying random stuff right now to see if something will work. But you are correct

PR #609, opened 2020-05-25
credview: quick fix for password glitch

dagheyman:
Sets currentDeviceValidated  to true for all devices that does not have OATH enabled, need to clean this up a bit later to make it more clear what it means.


fdennis:
Which pw glitch is this?

dagheyman:
When inserting a PW protected device, the "homeview" is visible for a second before the password prompt

PR #612, opened 2020-05-28
Interfaces to settings

fdennis:
TODO:

- [x] Call load() at the appropriate time, not as it is done currently

- [x] "Save interfaces" button should not always be enabled

- [x] Make load() more dynamic by using a list to iterate over

- [ ] Fix the rest of the checkboxes (FIDO, NFC, etc)

### DIFF
--- a/py/yubikey.py
+++ b/py/yubikey.py
@@ -12,8 +12,8 @@ from binascii import a2b_hex, b2a_hex
 from ykman.descriptor import (
     get_descriptors, list_devices, open_device,
     FailedOpeningDeviceException, Descriptor)
-from ykman.util import (TRANSPORT, APPLICATION, parse_b32_key)
-from ykman.device import YubiKey
+from ykman.util import (TRANSPORT, APPLICATION, Mode, parse_b32_key)
+from ykman.device import YubiKey, device_config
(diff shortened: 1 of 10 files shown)


py/yubikey.py
```diff
@@ -327,6 +327,42 @@ def refresh_devices(self, reader_filter=None):
...
+        with open_device(serial=self._current_serial) as dev:
+
+            try:
+                dev.write_config(
+                    device_config(
+                        usb_enabled=usb_enabled,
+                        nfc_enabled=nfc_enabled,
+                        ),
+                    reboot=True)
+            except APDUError as e:
```
dagheyman, line 348: If we are doing a version without supporting configuration lock, these specific error handling block should not be needed.

qml/utils.js
```diff
@@ -16,3 +16,37 @@ function hashCode(s) {
...
+ */
+function including(arr, value) {
+    if (includes(arr, value)) {
+        return arr;
+    } else {
+        return arr.concat(value);
+    }
+}
+
+function includes(arr, value) {
```
dagheyman, line 37: I think these are not needed, we should be able to use JS `includes` function directly, since we are on a newver version of Qt in this project

PR #618, opened 2020-06-15
format everything with qmlformat 

dagheyman:
Qt 5.15 comes with qmlformat, an automatic code formatter for qml.

This PR formats all qml with qmlformat and sets up a pre-commit hook as well.


dagheyman:
maybe not ready, seems to break the build by moving the callback functions to the top level

PR #621, opened 2020-06-22
Fixed snap not starting under Wayland

Toorero:
This patch allows the snap to be run under Wayland.

In addition, I fixed some errors I encountered during the build process.

### DIFF
--- a/docker/xenial-appimage/Dockerfile
+++ b/docker/xenial-appimage/Dockerfile
@@ -5,7 +5,7 @@ RUN apt-get update -qq \
     && add-apt-repository -y ppa:beineri/opt-qt-5.14.2-xenial \
     && apt-get -qq update \
     && apt-get -qq upgrade
-RUN apt-get install -y git make build-essential libssl-dev zlib1g-dev libbz2-dev \
+RUN apt-get install -y git make build-essential g++ libssl-dev zlib1g-dev libbz2-dev \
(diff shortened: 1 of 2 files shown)


dagheyman:
Thanks, this looks great. What OS and version are you testing this with?

docker/xenial-appimage/Dockerfile
```diff
@@ -5,7 +5,7 @@ RUN apt-get update -qq \
     && add-apt-repository -y ppa:beineri/opt-qt-5.14.2-xenial \
     && apt-get -qq update \
     && apt-get -qq upgrade
-RUN apt-get install -y git make build-essential libssl-dev zlib1g-dev libbz2-dev \
+RUN apt-get install -y git make build-essential g++ libssl-dev zlib1g-dev libbz2-dev \
```
dagheyman, line 8: This is a Dockerimage used to build the AppImage, unrelated to the snap. Did you mean to edit this?

Toorero:
To the first question: Ubuntu 20.04 LTS with the Ubuntu-Wayland desktop environment.

To the second point: If you say it's unrelated than undo it. I had some problems during the snap build process where "g++" was needed but not found. I will look into it. I would have to edit the build packages in the snapcraft.yaml then, right?

dagheyman:
Actually, there is no harm to keep it there as well. And the snap seems to build in the CI. Merging, thanks!

PR #638, opened 2020-06-30
shortcuts enabled in authenticator only

braathen:


qml/main.qml
```diff
@@ -271,31 +271,31 @@ ApplicationWindow {
...
+        enabled: isCurrentObjectName("authenticatorView") && !!currentCredentialCard
         onActivated: app.currentCredentialCard.calculateCard(true)
     }
 
     Shortcut {
         id: shortcutDelete
         property string description: "Delete account"
-        sequence: StandardKey.Delete
-        enabled: !!currentCredentialCard
+        sequence: Qt.platform.os === "osx" ? "Backspace" : "Delete"
```
dagheyman, line 281: Why?

braathen: StandardKey.Delete doesn't work on macOS for some reason.

PR #639, opened 2020-07-01
Actions: fix run installer step

fdennis:
Should fix actions problem on Win 64


fdennis:
Hmm, nope, does not fix it.

PR #661, opened 2020-09-17
New QR scanner

fdennis:


qml/NewCredentialView.qml
```diff
@@ -57,16 +60,38 @@ Flickable {
...
-        yubiKey.parseQr(ScreenShot.capture(), function (resp) {
-            scanning = false
-            if (resp.success) {
-                credential = resp
-            } else {
-                navigator.snackBarError(navigator.getErrorMessage(
-                                                                resp.error_id))
-            }
-        })
+        console.log("111")
```
nevun: 111oneone

fdennis: I think that is an outdated version of that file.

nevun: Ah, yes, you are right. Sorry about that :)

/me fetches more coffee

nevun (commented):
Would be good to document where and exactly which version of QZXing we pulled in. In case of bug/securitty fixes upstream for example.

fdennis:
> Would be good to document where and exactly which version of QZXing we pulled in. In case of bug/security fixes upstream for example (It seems it is [ftylitak/qzxing@4715b24](https://github.com/ftylitak/qzxing/commit/4715b242cc08582a0c31aad94d74251b77821a29) since the only diff between the pulled in code and upstream is the commit after that.)

You're right! Just pushed a fix for that.

PR #686, opened 2021-01-21
Set focus on search field when app is made visible, adds option to minimize window on code copy.

sarthurdev:
This PR adds a couple of improvements for quicker use. With these modifications you are able to open the app from the tray icon, instantly seach for the code you want, copy it to clipboard and enable the app to automatically hide again.

- Sets the default focus to search field when the app enters the foreground.
- Adds an option to minimize the window automatically after copying TOTP code.

### DIFF
--- a/qml/CredentialCard.qml
+++ b/qml/CredentialCard.qml
@@ -73,6 +73,10 @@ Pane {
     function copyCode(code) {
         clipBoard.push(code)
         navigator.snackBar(qsTr("Code copied to clipboard"))
+
+        if (settings.minimizeOnCopy) {
+            app.hide()
+        }
     }
 
     function calculateCard(copy) {
--- a/qml/SettingsView.qml
+++ b/qml/SettingsView.qml
@@ -566,6 +566,17 @@ Flickable {
                         indicator.height: 16
                         onCheckStateChanged: settings.hideOnLaunch = checked
                     }
+
+                    CheckBox {
+                        id: minimizeOnCopyCheckbox
+                        enabled: true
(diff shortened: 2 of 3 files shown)


braathen:
Thanks for the contribution! We have decided not to merge this, since it is designed for a specific workflow, and don't want to clutter the settings with too many options.

sarthurdev:
I don't see how this is a unique workflow, the sole purpose of this app is to access 2FA tokens and this PR just saves a few steps.

braathen:
From a UX perspective it's odd that things happens without the user explicitly asking for it, for example not everyone is using the search function to find their credentials, thus not expecting it to suddenly activate. Originally we had the search field automatically populate when you started typing, unfortunately we had to remove this due some people being confused when they tapped their yubikeys by mistake and it started searching for the OTP string (pleasing everyone is sometimes difficult and comes with tradeoffs).

You can quickly activate the search field by cmd/ctrl-f or moving the cursor up, you can copy the code by hitting return or cmd/ctrl-c when you've found it, and you can close the window with cmd/ctrl-w. We believe this to be the most common way of interaction.

Did you try the favorite option btw? If you favorite credentials they will appear in the menu bar/system tray for quick access without opening the window which also can be convenient in some cases.

sarthurdev:
The keyboard shortcuts do help, as do favourites for some of my colleagues. But some of us have 30+ tokens for various client systems which makes using favourites just as impractical.

Fair enough if you don't wish to pull this, I've made the build from this PR accessible to my team, I just figured that it may be useful to others.

PR #692, opened 2021-02-01
snap: prepare a revision for 5.0.4 to fix segfault problem

dagheyman:


dagheyman:
This seems to work but could use some more testing.
Tested on Ubuntu 20.10 so far.

dagheyman:
This seems to work but could use some more testing.
Tested on Ubuntu 20.10 so far.

dagheyman:
This seems to work but could use some more testing.
Tested on Ubuntu 20.10, 20.04 so far.

Issue #738, opened 2021-08-26
[ Request ] - Prompt User to enable smart card service (If disabled)


angry-bender:
- Yubico Authenticator version: 5.0.3, 5.0.5b, window store version, ykman-1.2.2
- Operating system and version: Windows 10 20H2
- YubiKey model and version: 
    - YubiKey 5 NFC (5.4.3) [OTP+FIDO+CCID]
     -YubiKey NEO (3.4.9) [OTP+CCID]
- Bug description summary: Windows 10 no longer recognises my the CCID/OATH interface in windows. This occurs in both the authenticator app and the ykman application. 

Steps to reproduce
---

Disable Smart Card Service

Run yubikey authenticator

Expected result
---

The application prompts the user (If other interfaces are detected, but not CCID/OATH) to enable the smart card service on a windows client

Actual results
---

Device isn't recognised / cant read CCID/OTP interface

Other info
---

Troubleshooting steps taken
#737 

Installed ykman-1.2.2
Ran  .\ykman.exe --log-level DEBUG oath info
Seen `smartcard.pcsc.PCSCExceptions.EstablishContextException: 'Failure to establish context: The Smart Card Resource Manager is not running. '`
Found the service was disabled
Enabled the service

dainnilsson:
Yubico Authenticator 6.0 has now been released and uses a new codebase. As such, this issue has been marked with the `legacy` label, and will be automatically closed in 7 days. If this issue is still relevant to Yubico Authenticator 6, please comment on the issue saying so, and it will be kept open (or be re-opened). Sorry for the inconvenience!

PR #739, opened 2021-08-26
Added smart card service prompt from Issue #738

angry-bender:
Added a draft message to prompt the user to enable the smart card service.

recommended improvements

1. Check for windows environment first
2. Check if other non OATH interfaces are detected first


dainnilsson:
Thank you for your contribution! This PR is no longer applicable to the current code base (version 6.0), but I believe this case should be handled better there already, so hopefully it is no longer needed. If you think it is, please open a new issue or PR.

PR #756, opened 2021-11-06
snap: ensure the pcscd daemon is up

freedge:
snap: ensure the pcscd daemon is up

After some inactivity, the pcscd daemon will gracefully shutdown, preventing
yubioath to detect the key. Users are instructed to restart pcscd by hand when
that happens.

Fix it: ensure the daemon does not shutdown, and is restarted when that
happen.

Resolves: https://github.com/Yubico/yubioath-desktop/issues/656

### DIFF
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -7,8 +7,10 @@ confinement: strict
 
 apps:
   pcscd:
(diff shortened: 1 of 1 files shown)


arkel-s:
That will be helpful!

fdennis:
Thank you for the help! Will merge.

PR #781, opened 2022-02-12
Set QGuiApplication::desktopFileName

nicolasfella:
This is needed for proper window identification on Wayland


gbcox:
I've tested this on Fedora 35 and it resolves the wayland titlebar issue.  Much appreciated.  If this could be rolled into the next update, that would be appreciated.  Otherwise, I'll patch the Fedora package.

dainnilsson:
Thank you for your contribution! This PR is no longer applicable to the current code base (version 6.0). I'm not sure if this is still an issue with the new version or not. If you think it is, please open a new issue or PR.

PR #782, opened 2022-02-12
Mark as single window in desktop file

nicolasfella:
Only one instance can be active at a time, indicate that in the desktop file

### DIFF
--- a/resources/com.yubico.yubioath.desktop
+++ b/resources/com.yubico.yubioath.desktop
@@ -9,3 +9,5 @@ Terminal=false
 Type=Application
 Categories=Utility;
 Keywords=Yubico;Authenticator;
+SingleMainWindow=true
+Version=1.5


gbcox:
I tried to build this patch and it failed on desktop-file-utils validation.  Apparently, even though the current docs say that the version is 1.5, they haven't yet updated the desktop-utils yet, so the build fails.  There is a bug:  https://gitlab.freedesktop.org/xdg/desktop-file-utils/-/issues/59 that was entered 9 months ago complaining about this, but as of yet, no action.  I also entered a bug here:  https://bugzilla.redhat.com/show_bug.cgi?id=2053905
I guess what I'm trying to say is you shouldn't merge this until this situation gets resolved.  The other patch (781) works perfectly.  I'm able to see the correct icon now in the wayland titlebar.

gbcox:
Upstream patch has been merged:   https://gitlab.freedesktop.org/xdg/desktop-file-utils/-/issues/59
Fedora has applied and pushed to repositories, and other distros now have the capability to update.  This is good to go IMO.

dainnilsson:
Thank you for your contribution! This PR is no longer applicable to the current code base (version 6.0). We have several changes to the desktop file in the new version and we believe it acts correctly already, without these changes. If you find that not to be the case (or find new issues with it!), please open a new issue/PR against the new codebase.

PR #824, opened 2022-11-18
Fix Windows build

dainnilsson:
The Windows build started failing due to paths being too long. This changes some work directories to be shorter.


AdamVe (approved):
LGTM

PR #825, opened 2022-11-18
macOS: release script for standalone app

fdennis:


dainnilsson:
Maybe add some error checking at the start? Like make sure username and password are given, check if any required executables are installed (I think it requires something for the dmg right?), etc.

dainnilsson (approved):

PR #827, opened 2022-11-21
[Android] fix crash when refreshing oath credentials

AdamVe:
- Fixes crashes on Android when memory keys for OathSession have expired and we try to execute methods on locked oath session.


AdamVe:
replaced by #829

PR #828, opened 2022-11-21
[Android] handle locked oath sessions when refreshing

AdamVe:
When using YubiKey with a password protected Oath applet, the users have to provide the Oath password to be able to access the applet. This information is cached in a memory and is automatically cleared up after a delay (unless the users chooses "Remember password" when unlocking the Oath applet).

This PR fixes issue, when the application crashed because of trying to execute a method after the cached unlocking key already expired.

Note: Other methods on the oath session are executed through methodChannel handler and are caught/handled in flutter space.


AdamVe:
replaced by #829

PR #847, opened 2022-11-25
Detect a non-working helper and show it in the UI.

dainnilsson:
This detects a failure to communicate with the Helper (using a timeout of 5 seconds) and puts the app in an error state, making it more obvious that it has failed:
image


AdamVe (approved):
LGTM

PR #857, opened 2022-11-30
Fix release scripting for macos and win

fdennis:
For a standalone app we need two arguments in order to notarize. For the app store we do not need anything at all.


AdamVe (approved):
LGTM

PR #868, opened 2022-12-08
[Linux] Build Flutter app in container

dainnilsson:
The Flutter app also becomes dependent on the glibc version available on the machine, so we need to build it on an older OS. Since the ubuntu-18.04 runner is going away very soon, we use a Docker container for this.


fdennis:
Tested - works!

fdennis (approved):

PR #880, opened 2022-12-20
Fix waiting for Helper timeout

dainnilsson:
This changes how we wait for the Helper process, indicating that it is slow after a sufficient delay, but still recovering after the delay if it becomes responsive.
This also introduces delayed visibility of AppPage, which is used to delay the showing of the CircularProgressIndicators when the page is only displayed for a split second.


AdamVe (approved):
Improves user experience. Tested on Mac OS and Android and did not encountered any issues.

lib/app/views/device_error_screen.dart
```diff
@@ -48,7 +48,7 @@ class DeviceErrorScreen extends ConsumerWidget {
                     context, 'Elevating permissions...',
                     duration: const Duration(seconds: 30));
                 try {
-                  if (await ref.read(rpcProvider).elevate()) {
+                  if (await ref.read(rpcProvider).value!.elevate()) {
```
fdennis, line 51: Is there a reason for having `value!` here instead of `requireValue`?

dainnilsson: There isn't, good catch!

lib/desktop/init.dart
```diff
@@ -145,38 +141,42 @@ Future<Widget> initialize(List<String> argv) async {
...
       )
     ],
     child: YubicoAuthenticatorApp(
       page: Consumer(
-        builder: ((_, ref, child) {
+        builder: ((context, ref, child) {
           // keep RPC log level in sync with app
           ref.listen<Level>(logLevelProvider, (_, level) {
-            rpc.setLogLevel(level);
+            ref.read(rpcProvider).valueOrNull?.setLogLevel(level);
```
fdennis, line 155: I'm also wondering if this should be changed to `requireValue` as well?

dainnilsson: No, in this case it is correct. We don't want to fail if this happens before the rpcProvider is initialized, so null here is fine (we'll just not call setLogLevel). `requireValue` would thrown an exception in that case.

fdennis (approved):

lib/desktop/init.dart
fdennis: Alright, cool!

PR #902, opened 2023-01-10
Explicitly set utf-8 for stdin/out/err.

dainnilsson:
This was causing problems on Windows when sending certain non-ascii characters.


AdamVe (approved):
LGTM

PR #913, opened 2023-01-17
[Android] handle exceptions when creating connections

AdamVe:
When `requestConnection` has a problem, it is stored in `Result<C, IOException>` and will be thrown by accessing the `value` property (see [Result.java](https://github.com/Yubico/yubikit-android/blob/2.1.0/core/src/main/java/com/yubico/yubikit/core/util/Result.java#L51)). `runCatching` would swallow all these exceptions. 

This is fixed in this PR by resuming the continuation with the exceptions caught by `runCatching` and clients of `withConnection` can do the handling appropriately.


AdamVe:
I reviewed this again and `runCatching` does not swallow the exceptions from `requestConnection`. This PR is obsolete.

PR #919, opened 2023-01-20
[Android] handle unrecognized NFC devices

AdamVe:
When an NFC device is used and the app cannot recognize it, it will be shown on the main page.

### DIFF
--- a/lib/app/views/main_page.dart
+++ b/lib/app/views/main_page.dart
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2022 Yubico.
+ * Copyright (C) 2022-2023 Yubico.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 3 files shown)


lib/app/views/main_page.dart
```diff
@@ -113,7 +113,12 @@ class MainPage extends ConsumerWidget {
       return ref.watch(currentDeviceDataProvider).when(
             data: (data) {
               final app = ref.watch(currentAppProvider);
-              if (app.getAvailability(data) == Availability.unsupported) {
+              if (data.info.config.enabledCapabilities.isEmpty &&
```
fdennis, line 116: Why do you need the `data.info.config.enabledCapabilities.isEmpty` check? Is it not enough to simply check the name only?

AdamVe: I decided to do it this way because I know what the `data` looks like for unrecognized devices, when that situation happens - it is constructed by [this code](https://github.com/Yubico/yubioath-flutter/pull/919/files#diff-1defc8f773208b04e321468e9a2c644111b109a46d1cbb344ccd314e020a3c78R304-R321)

fdennis (approved):

lib/app/views/main_page.dart
dainnilsson: It might be cleaner to check `data.info.supportedCapabilities` since that data is static for a key, whereas `enabledCapabilities` can be modified by the user.

dainnilsson (approved):

lib/app/views/main_page.dart
AdamVe: I see - I will rather use supported capabilities, thanks

PR #933, opened 2023-01-26
Use Flutter 3.7.0

dainnilsson:
This breaks a lot of the styles for now, so we'll need to made additional adjustments.


AdamVe (approved):
Looks good!

PR #940, opened 2023-01-31
Add CodeQL

AdamVe:
Adds CodeQL action based on action from the `legacy` branch. Supported languages are `kotlin` and `python`

### DIFF
--- a/.github/workflows/codeql-analysis.yaml
+++ b/.github/workflows/codeql-analysis.yaml
@@ -56,17 +56,10 @@ jobs:
           YUBIOATH_STORE_PASSWORD: ${{ secrets.YUBIOATH_STORE_PASSWORD }}
         working-directory: ./app
 
-      - if: matrix.language == 'kotlin'
-        name: Perform CodeQL Analysis
-        uses: github/codeql-action/analyze@v2
 
       - if: matrix.language == 'python'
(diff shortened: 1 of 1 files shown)


jpreese:
This LGTM @AdamVe 🎉  Thanks for adding CodeQL!

The only suggestion I would have is that you could add some specific folder paths and/or branches so that the action does not run on changes that won't contain any Python or Kotlin files. We have a pool of minutes that are consumed when any GitHub Action runs, so anything helps!

You can see an example of this syntax here: https://github.com/Yubico/yubienterprise-services/blob/main/.github/workflows/codeql.yaml#L3

dainnilsson:
> The only suggestion I would have is that you could add some specific folder paths and/or branches so that the action does not run on changes that won't contain any Python or Kotlin files. We have a pool of minutes that are consumed when any GitHub Action runs, so anything helps!

Why does it use paid minutes? Does it trigger some workflow in another repo or something?

AdamVe:
Added the suggested improvements to trigger only if a change occurs in directories which contain python or java/kotlin code. 
Tested this in another branch and it works as expected, only changed inside `android/` or `helper/` will trigger the code ql workflow.

dainnilsson (approved):

PR #942, opened 2023-02-01
Fix linux library path

dainnilsson:
When grabbing screenshots using gnome-screenshot we need to reset the LD_LIBRARY_PATH.

Also adds support for using spectacle (on KDE).


AdamVe (approved):
LGTM

PR #955, opened 2023-02-08
Handle state of NFC adapter

AdamVe:
Makes the state of NFC adapter explicit through the app.
- devices without NFC adapter will not ask to "Tap" the YubiKey
- when the NFC adapter is off, it can be turned on from the main screen
- tested on mobile and Chromebook devices

### DIFF
--- a/lib/app/views/main_page.dart
+++ b/lib/app/views/main_page.dart
@@ -16,6 +16,8 @@
 
 import 'package:flutter/material.dart';
 import 'package:flutter_riverpod/flutter_riverpod.dart';
+import 'package:yubico_authenticator/android/app_methods.dart';
+import 'package:yubico_authenticator/android/state.dart';
 
 import '../../exception/cancellation_exception.dart';
 import '../../core/state.dart';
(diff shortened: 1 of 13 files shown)


lib/app/views/main_page.dart
```diff
@@ -67,9 +74,27 @@ class MainPage extends ConsumerWidget {
...
+
+        var hasNfcSupport = ref.watch(androidNfcSupportProvider);
+        var isNfcEnabled = ref.watch(androidNfcStateProvider);
         return MessagePage(
           graphic: noKeyImage,
-          message: 'Insert or tap your YubiKey',
+          message: hasNfcSupport
+              ? isNfcEnabled
+                ? const Text('Tap or insert your YubiKey')
+                : Column(
```
dainnilsson, line 85: Instead of changing `message` into a Widget, the `actions` member of `MessagePage` should be used for the button. There's an example of this in `app_failure_page.dart`

dainnilsson (changes requested):
Instead of changing `message` from a string into a Widget, the `actions` member of `MessagePage` should be used for the button. There's an example of this in `app_failure_page.dart`. This ensures consistency between the different pages.

dainnilsson (approved):

PR #963, opened 2023-02-13
Alternative OATH navigation

dainnilsson:
This turns the "code pill" into its own button that opens the dialog view. Single click on the list item itself selects it, double-click copies code to clipboard (this was previously on the "pill"). For Android the tap/longpress behavior remains unchanged.


lib/oath/views/account_helper.dart
```diff
@@ -0,0 +1,155 @@
...
+      return '';
+    } else if (value.length < 6) {
+      return value;
+    } else {
+      var i = value.length ~/ 2;
+      return '${value.substring(0, i)} ${value.substring(i)}';
+    }
+  }
+
+  Widget buildCodeLabel() => Opacity(
```
AdamVe, line 141: I would recommend changing this function into a stateless widget.

AdamVe (approved):
tested on Android and it works good 👍🏻

PR #968, opened 2023-02-22
Icon pack support

AdamVe:
### Implements support for icon packs.

Account avatars will use icons from the icon pack if available.

Currently, the following icon packs have been tested:
- https://aegis-icons.github.io/
- https://github.com/alexbakker/aegis-simple-icons

### More information:
- only one icon pack can be loaded at a time
- tested with svg, png and jpg icons (various sizes)
- issuer name is not case sensitive
- the first match from the icon pack will be used
- the icon pack is copied to the application's support directory and used icons are cached for better performance. The files can be located in the following directoris:
  - macOS: `~/Library/Containers/com.yubico.yubioath/Data/Library/Application Support/com.yubico.yubioath`
  - Windows: `%homedrive%%homepath%\AppData\Roaming\Yubico\Yubico Authenticator`
  - Linux: `~/.local/share/authenticator`
  - Android: `/data/data/com.yubico.yubioath/files`

- icon pack archive file size is limited to 5MB

### Usage instructions
1. In the OATH menu select Custom icons:
image

2. Click `Load icon pack` and select the archive (.zip)
image

3. The icons for account which have icon in the icon pack will be updated automatically
image

4. To change loaded icon pack or remove the currently loaded, open the Custom icons dialog from the OATH menu and choose 'Replace icon pack' to choose a new pack or the tap the delete button to remove the current icon pack:
image


dainnilsson (commented):
If we're adding the `path` dependency we should make better use of it. We shouldn't manualy concatenate strings or use Platform.pathSeparator when there is a `join` for this purpose.

AdamVe:
> If we're adding the `path` dependency we should make better use of it. We shouldn't manualy concatenate strings or use Platform.pathSeparator when there is a `join` for this purpose.

You are right. I missed the `join` method. This will be updated.

lib/l10n/app_en.arb
```diff
@@ -81,6 +81,28 @@
             "capacity": {}
         }
     },
+    "oath_custom_icons": "Custom icons",
+    "oath_custom_icons_description": "Icon packs can make your accounts more easily distinguishable with familiar logos and colors. ",
```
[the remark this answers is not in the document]
AdamVe: It was intended as this sentence is composed in a text span and followed by next string. I will try to create the space programatically.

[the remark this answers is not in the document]
AdamVe: Fixed now

lib/l10n/app_en.arb
```diff
@@ -81,6 +81,28 @@
             "capacity": {}
         }
     },
+    "oath_custom_icons": "Custom icons",
+    "oath_custom_icons_description": "Icon packs can make your accounts more easily distinguishable with familiar logos and colors.",
+    "oath_custom_icons_replace": "Replace icon pack",
+    "oath_custom_icons_loading": "Icon pack is loading...",
+    "oath_custom_icons_load": "Load icon pack",
```
fdennis, line 88: Maybe it's worth pointing out somewhere that we are expecting a .zip file? For example perhaps just `Load icon pack (.zip)` would do fine?

braathen: Maybe, but is it really needed? The file dialog will only show .zip files as selectable, but if we're thinking about the flow you first downloaded the icon package from somewhere, it doesn't really matter what the extension is, you downloaded a file and that's the one you're going to be looking for.

dainnilsson: I think the filter in the file picker is probably sufficient. I would change "Icon pack is loading..." to "Loading icon pack..." though.

AdamVe: I changed the _Loading_ copy to suggested string and kept the _Load_ copy without the (.zip).

dainnilsson (approved):

PR #969, opened 2023-02-22
Persist window position

AdamVe:
After the app is reopened, the window will be placed to the last known position. 
If the display configuration changed (for example a monitor was added/removed or any scaling has been changed) the window will be placed to default location - the goal is to make sure that the window is never lost.

Currently only available for MacOS and Windows.


AdamVe:
replaced by #982

PR #980, opened 2023-03-01
Refactor external strings

dainnilsson:
This PR reorganizes the external strings used by the app, as well as extracts most of the remaining hard-coded values.
Strings have been renamed, deduplicated, and reordered.


lib/l10n/app_en.arb
```diff
@@ -1,301 +1,430 @@
...
+    "l_insert_or_tap_yk": "Insert or tap a YubiKey",
+    "l_unplug_yk": "Unplug your YubiKey",
+    "l_reinsert_yk": "Reinsert your YubiKey",
+    "l_place_on_nfc_reader": "Place your YubiKey on the NFC reader",
+    "l_replace_yk_on_reader": "Place your YubiKey back on the reader",
+    "l_remove_yk_from_reader": "Remove your YubiKey from the NFC reader",
+    "l_try_reinsert_yk": "Try to remove and reinsert your YubiKey.",
+    "l_touch_required": "Touch Required",
+    "l_touch_button_now": "Touch your YubiKey now",
+    "l_touch_button_now": "Touch the button on your YubiKey now",
```
AdamVe, line 105: it looks like this key is twice here

AdamVe: On Android we originally used `Touch the button on your YubiKey now.`

dainnilsson: Good catch! We can probably choose one and use it for both platforms.

lib/l10n/app_en.arb
```diff
@@ -1,301 +1,430 @@
...
-    "widgets_close": "Close",
+    "@_yubikey_interactions": {},
+    "l_insert_yk": "Insert your YubiKey",
+    "l_insert_or_tap_yk": "Insert or tap a YubiKey",
+    "l_unplug_yk": "Unplug your YubiKey",
+    "l_reinsert_yk": "Reinsert your YubiKey",
+    "l_place_on_nfc_reader": "Place your YubiKey on the NFC reader",
+    "l_replace_yk_on_reader": "Place your YubiKey back on the reader",
+    "l_remove_yk_from_reader": "Remove your YubiKey from the NFC reader",
+    "l_try_reinsert_yk": "Try to remove and reinsert your YubiKey.",
```
AdamVe, line 102: note: in original it is spelled re-insert, but reinsert is correct.

lib/l10n/app_en.arb
```diff
@@ -1,301 +1,430 @@
...
+    "l_add_fingerprint": "Add fingerprint",
+    "l_fp_step_1_capture": "Step 1/2: Capture fingerprint",
+    "l_fp_step_2_name": "Step 2/2: Name fingerprint",
+    "l_delete_fingerprint": "Delete fingerprint",
+    "l_fingerprint_deleted": "Fingerprint deleted",
+    "l_warning_delete_fingerprint": "This will delete the fingerprint from your YubiKey.",
+    "l_no_fingerprints": "No fingerprints",
+    "l_set_pin_fingerprints": "Set a PIN to register fingerprints",
+    "l_no_fps_added": "No fingerprints have been added",
+    "l_fingerprint_renamed": "Fingerprint renamed",
```
AdamVe, line 313: the naming of the keys for `fingerprint` sometimes uses `fp`/`fps`. I think it would be nice to unify this, so one can easily find all fingerprint related UI strings by search for a one string (either use only fp or only fingerprint)

lib/oath/keys.dart
```diff
@@ -18,7 +18,7 @@ import 'package:flutter/material.dart';
 
 const _prefix = 'oath.keys';
 
-const setOrManagePasswordAction = Key('$_prefix.set_or_manage_oath_password');
+const setOrManagePasswordAction = Key('$_prefix.set_or_manage_w_password');
```
AdamVe, line 21: is this change intended? should be better with `set_or_manage_password`. note that this is not a localization key but a string for Widget key.

AdamVe (approved):
👍🏻

Issue #882, opened 2022-12-27
App does not remember window position


edelstone:
Problem: I like using the application in a narrow size, justified all the way to the left of my screen. However, after quitting and restarting the app, it always moves towards the center of the screen instead of remaining justified to the side. It's a minor annoyance that was actually fixed at one point in a previous version (#574).

Apple M1 Pro 2021 16" MBP
MacOS Ventura 13.1
Version 6.0.2 (downloaded from Mac App Store)
Yubikey 5C Nano

GitPullNow:
Have the same issue starting from v6. Previous version 5 did not have this issue for me. It does not remember on which screen it is or what size the window should be at.

PC Windows 11 22H2 Build 22621.1194
Dual monitor setup with a 24" (1920x1080) and a 34" Ultrawide (3440x1440)
Yubico Authenticator v6.1.0
Yubikey 5C Nano

AdamVe:
Hi! We have merged support for remembering window position of the app. Currently only **MacOS** and **Windows** platforms are supported.

It is possible to try out the feature by running the build from GitHub. If you do, please, let us know what you think.

edelstone:
Using Production version 6.2.0 and this is working great. Thank you.

PR #982, opened 2023-03-02
Persist window position (v2)

AdamVe:
After the app is reopened, the window will be placed to the last known position.
If the display configuration changed (for example a monitor was added/removed or any scaling has been changed) the window will be placed to default location - the goal is to make sure that the window is never lost.

Currently only available for MacOS and Windows.

This is a new iteration of #969 with simpler code:
- rebased on main with System Tray support
- no extra plugin or native code


lib/desktop/window_manager_helper/_wm_helper_windows_impl.dart
```diff
@@ -0,0 +1,96 @@
...
+  }
+
+  static Future<bool> _displayConfigurationChanged(
+      SharedPreferences sharedPreferences) async {
+    final allDisplays =
+        sharedPreferences.get(_keyAllDisplaysValue) as String? ?? '';
+    return await _getAllDisplays() != allDisplays;
+  }
+
+  static bool displayContainsBounds(Display d, Rect rect) {
```
AdamVe, line 42: change name to `_displayContainsBounds`

dainnilsson (approved):

AdamVe:
Related to #882

PR #998, opened 2023-03-08
a11y: read issuer, name and code when using VO

fdennis:
This is a small PR that basically creates a semantics label for the credentials. The label will contain issuer, name, and the code. With both Narrator (Win) and VoiceOver (macOs) the credentials are read correctly with this fix.

### DIFF
--- a/lib/oath/views/account_view.dart
+++ b/lib/oath/views/account_view.dart
@@ -99,6 +99,24 @@ class _AccountViewState extends ConsumerState<AccountView> {
     }).toList();
   }
 
+  String? a11yCredentialLabel(String? issuer, String? name, OathCode? code) {
+    String? label = '';
+    String? tmpIssuer = issuer;
+    String? tmpName = name;
+    String? tmpCode = code?.value;
+    if (tmpIssuer != null) {
+      label += tmpIssuer;
+    }
+    if (tmpName != null) {
+      label += tmpName;
+    }
+    if (tmpCode != null) {
+      label += tmpCode;
+    }
+
+    return label;
+  }
+
   @override
   Widget build(BuildContext context) {
     final theme = Theme.of(context);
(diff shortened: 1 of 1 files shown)


lib/oath/views/account_view.dart
```diff
@@ -99,6 +99,24 @@ class _AccountViewState extends ConsumerState<AccountView> {
...
+    if (tmpName != null) {
+      label += tmpName;
+    }
+    if (tmpCode != null) {
+      label += tmpCode;
+    }
+
+    return label;
+  }
+
```
dainnilsson, line 119: A lot of the `String?` types here don't need to be nullable. I'd change the return type, as well as the type of `name` and `label` to `String`. I also don't think any of the `tmp` variables are needed here.

lib/oath/views/account_view.dart
```diff
@@ -99,6 +99,24 @@ class _AccountViewState extends ConsumerState<AccountView> {
     }).toList();
   }
 
+  String? a11yCredentialLabel(String? issuer, String? name, OathCode? code) {
```
dainnilsson, line 102: This should probably be a private function (rename to `_ally...`) and doesn't need to be part of the class, I'd move it out.

lib/oath/views/account_view.dart
```diff
@@ -99,6 +99,24 @@ class _AccountViewState extends ConsumerState<AccountView> {
...
+    String? tmpCode = code?.value;
+    if (tmpIssuer != null) {
+      label += tmpIssuer;
+    }
+    if (tmpName != null) {
+      label += tmpName;
+    }
+    if (tmpCode != null) {
+      label += tmpCode;
+    }
```
dainnilsson, line 115: Don't we need spaces between the parts for this to be read out correctly?

dainnilsson (changes requested):

lib/oath/views/account_view.dart
fdennis: Agreed!

lib/oath/views/account_view.dart
fdennis: Good catch. Absolutely. Since account name is required I will not add a space to it, but instead add a space before issuer and code.

dainnilsson (approved):

PR #1009, opened 2023-03-14
Add auto fill hints to secret input fields

AdamVe:
This is a workaround recommended in https://github.com/flutter/flutter/issues/122300.

Will be effective with future flutter release.

This change also partially enables use of an Autofill Service on Android - currently we can only fill in saved passwords. Saving/Updating new passwords does not work.

### DIFF
--- a/lib/oath/views/add_account_page.dart
+++ b/lib/oath/views/add_account_page.dart
@@ -453,6 +453,7 @@ class _OathAddAccountPageState extends ConsumerState<OathAddAccountPage> {
                       key: keys.secretField,
                       controller: _secretController,
                       obscureText: _isObscure,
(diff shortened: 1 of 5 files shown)


lib/oath/views/add_account_page.dart
```diff
@@ -453,6 +453,7 @@ class _OathAddAccountPageState extends ConsumerState<OathAddAccountPage> {
                       key: keys.secretField,
                       controller: _secretController,
                       obscureText: _isObscure,
+                      autofillHints: isAndroid ? [] : const [AutofillHints.password],
```
dainnilsson, line 456: This took me a while to realize why it was only for non-Android. Maybe adding a comment would make it more clear.

dainnilsson (approved):

PR #1036, opened 2023-04-13
Call `tryToUnlockOathSession` conditionally

AdamVe:
The call `tryToUnlockOathSession()` is used for every OATH action: if possible, the session is unlocked and the action can complete. 

This is not ideal for `unsetPassword` where we need to have the session locked to verify that the user who is unsetting the password knows the password.

This PR adds a new parameter `unlock: Boolean` (defaults to true) to `useOathSession` call. The session is automatically unlocked only if the parameter value is true. `unsetPassword` explicitly sets the value to false.


dainnilsson (approved):
LGTM

PR #1050, opened 2023-04-18
Fix to Oath unlock when password is "remembered"

AdamVe:
This PR fixes following issue:

1. set password on Oath Session
2. restart app, connect YubiKey (USB or NFC) -> app asks for password
3. check "Remember password", type in password, tap "Unlock"
4. restart app, connect YubiKey (USB or NFC)

Expected: no password asked
What really happened: app asks for password

The issue was introduced int #1036 where we explicitly disabled automatic unlock for "unsetPassword". This introduced the described erroneous behaviour, because we moved the automatic unlock away from the `device.withConnection()` calls and the session was not unlocked for the "remembered" cases.

The fix is that we unlock conditionally directly after `device.withConnection()` (both USB and NFC). The condition is set through `useOathSession()` call so we can have this per feature (unsetPassword etc)

This PR also disables the automatic unlock for `setPassword` which is used also when changing passwords and had the same issue as `unsetPassword` (see #1036).

This was tested manually.

### DIFF
--- a/android/app/src/main/kotlin/com/yubico/authenticator/oath/OathManager.kt
+++ b/android/app/src/main/kotlin/com/yubico/authenticator/oath/OathManager.kt
@@ -63,6 +63,7 @@ import kotlinx.coroutines.*
 import kotlinx.serialization.encodeToString
 import java.net.URI
 import java.util.concurrent.Executors
+import java.util.concurrent.atomic.AtomicBoolean
 import kotlin.coroutines.suspendCoroutine
 
 typealias OathAction = (Result<YubiKitOathSession, Exception>) -> Unit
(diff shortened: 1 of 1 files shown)


android/app/src/main/kotlin/com/yubico/authenticator/oath/OathManager.kt
```diff
@@ -100,6 +101,7 @@ class OathManager(
     @TargetApi(Build.VERSION_CODES.M)
     private fun createKeyStoreProviderM(): KeyProvider = KeyStoreProvider()
 
+    private var unlockOnConnect = AtomicBoolean(true)
```
dainnilsson, line 104: This can probably be `val`.

dainnilsson (approved):

Issue #1051, opened 2023-04-19
Loading icon pack fails with Unhandled Exception: FileSystemException: Rename failed


jonasbb:
I see this error message on the command line when trying to load an icon pack. The app is run in flatpak (https://github.com/flathub/com.yubico.yubioath/pull/69) on a Fedora host. I downloaded the Aegis Icon pack from https://aegis-icons.github.io/.

```
[ERROR:flutter/runtime/dart_vm_initializer.cc(41)] Unhandled Exception: FileSystemException: Rename failed, path = '/tmp/yubioathCHZTQM/unpack' (OS Error: Invalid cross-device link, errno = 18)
#0      _checkForErrorResponse (dart:io/common.dart:42)
#1      _Directory.rename.<anonymous closure> (dart:io/directory_impl.dart:200)
<asynchronous suspension>
#2      IconPackManager.importPack (package:yubico_authenticator/oath/icon_provider/icon_pack_manager.dart:169)
<asynchronous suspension>
#3      _ImportActionChip._importAction (package:yubico_authenticator/oath/icon_provider/icon_pack_dialog.dart:183)
<asynchronous suspension>
```

The UI is stuck showing "Loading icon pack...` and spinning non-stop.

The error indicates that the app is trying to perform a filesystem rename across different filesystems. `/tmp` is a tmpfs. Either this needs to be checked or the code should always copy the files.

AdamVe:
We merged the fix provided by @evan-a-a, it is now available in main and will be part of the next release of the app. Thanks for reporting!

PR #1052, opened 2023-04-19
icons: Avoid use of rename when installing icon pack

evan-a-a:
rename does not work across filesystems. /tmp is often in a tmpfs filesystem, and thus installing the icon pack fails.

Solve this by using a copy/delete pattern instead of rename

Fixes https://github.com/Yubico/yubioath-flutter/issues/1051


AdamVe:
Thank you for the fix @evan-a-a, I will merge this PR asap.

AdamVe (approved):
👍🏻 Thanks!

Issue #773, opened 2022-01-18
Can't read QR-codes in wayland


traffic-light:
- Yubico Authenticator version: V5.1.0
- Operating system and version: Arch linux with KDE, using wayland as window magager
- YubiKey model and version: Yubikey 5 nfc
- Bug description summary: Can't read QR-code from Xwayland window

The application runs in wayland but firefox and chrome runs in Xwayland (X11 server of wayland)

I don't think there is an easy (if there is) fix for it. But I wanted to let you guys know.

(didn't find a dirty fix to force run it in Xwayland...)

JonnieCache did look further into it. His post:
> In my testing I've found that yubico a
> uthenticator runs under xwayland, and so does chrome. Firefox runs under native wayland. QR code scanning doesnt work from either browser.
> 
> Firefox is 97.0, Chrome is 98.0.4758.80, running ubuntu 21.10.

JonnieCache:
In my testing I've found that yubico authenticator runs under xwayland, and so does chrome. Firefox runs under native wayland. QR code scanning doesnt work from either browser.

Firefox is 97.0, Chrome is 98.0.4758.80, running ubuntu 21.10.

fdennis:
@traffic-light Is the issue solved by that reply? If so, can this be closed?

traffic-light:
I don't think it's really resolved. There is no solution ;-)

But I think it's just a specific windowmanager issue...

JonnieCache:
The app needs to be updated to use the new wayland screen capture api, something like [this](https://flatpak.github.io/xdg-desktop-portal/#gdbus-org.freedesktop.portal.Screenshot)

beyond my ken I'm afraid.

alienscience:
I just hit this problem and used the following workaround on a site that only had the qr code and did not display the secret key:

- Save the image to a file using the web browser.
- Use the `zbarimg` program that is in most linux repositories to print the contents of the qr code.

dainnilsson:
Yubico Authenticator 6.0 has now been released and uses a new codebase. As such, this issue has been marked with the `legacy` label, and will be automatically closed in 7 days. If this issue is still relevant to Yubico Authenticator 6, please comment on the issue saying so, and it will be kept open (or be re-opened). Sorry for the inconvenience!

schmitmd:
I don't believe this works under Yubico Authenticator 6.

dainnilsson:
[...]

PR #1053, opened 2023-04-19
Simplify screenshot capture and add support for wayland

### LINKED ISSUE #1010: Use XDG Desktop Portals on Linux instead of screenshot tools
The application needs an ability to scan QR code from screen, but arbitrary reading from screen is not supported on Wayland for security reasons. But an ability to get screen content when needed is needed sometimes. So there is an API for that. There is a generic API for permission-based access granting on Linux. It's called [XDG Desktop Portals](https://github.com/flatpak/xdg-desktop-portal). There is [a portal](https://flatpak.github.io/xdg-desktop-portal/#gdbus-org.freedesktop.portal.Screenshot) for taking screenshot.

It's already used in some apps for the same thing. For example, [Authenticator](https://gitlab.gnome.org/World/Authenticator), an application for generating two-factor codes, [uses the portal](https://gitlab.gnome.org/World/Authenticator/-/blob/master/src/widgets/camera.rs#L43) to scan QR code.

There are some advantages of this way to scan QRs from the screen:
- **Universality.** The portal will work on every desktop environment that support it. Your current hack using GNOME Screenshot and Spectacle requires GNOME Screenshot and Spectacle to be installed and also only works on GNOME and KDE, but the portal works on GNOME, KDE, Elementary, wlroots-based WMs and every desktop environment that supports or will support the portal.
- **Sandboxability.** The portal works well with sandboxing systems like Flatpak and Snap, so the application will be able to work without unrestricted access to the host system.
- **Consistency.** You won't need to use different parts of code for different Linux desktop environments since the portal API is unified.
- **Availability.** GNOME Screenshot and Spectacle can be or not bet installed on the machine. But XDG Desktop Portals is very common and standard thing that is usually installed on every Linux computer.
- **Long-term support**: Your current hack with GNOME Screenshot and Spectacle supposes that these tools won't change their CLI and behavior. But they can, since they're not designed as any kind of public API. The portal API *is* a public API that is expected to be stable and supported since applications are expected to use it.

evan-a-a:
pyscreenshot supports capturing screenshots via numerous backends and can automatically select the best one depending on platform. This enables capturing of the QR code on wayland-based desktops while maintaining support for Windows.

Fixes #1010


evan-a-a:
This should be working now, at least on Linux. I'll give it a checkout on Windows once the test build is done. I don't have a macOS machine available to test.

Small note, I tested this in the Flatpak but there's a small bug in `pyscreenshot` that causes an issue. I am going to patch pyscreenshot to fix the behavior, but I don't think this should block merging this request. The dependency can be updated later.

fdennis:
Hi, and thanks for the PR!
We have tried the builds on win10, win11, and macOS. Unfortunately, scanning does not work properly. When pressing "Scan QR code" the spinner spins and never stops. Are you able to reproduce this on your Windows?

trinitronx:
I tested merging this with latest `main` @ e753fac5 on Manjaro Sway with Wayland `1.22.0`, and `gnome-screenshot` `41.0+r25+g45f08f0-1`.  Unfortunately, I was also able to [reproduce the spinner bug][1] mentioned by @fdennis.  It appears that at least on Wayland `wlroots`-based compositors (e.g. without X11, nor Gnome Shell + Mutter (as compositor), `gnome-screenshot` does not support screen capture.

Upstream bug report is here: [GNOME/gnome-screenshot#201](https://gitlab.gnome.org/GNOME/gnome-screenshot/-/issues/201). (It was closed as "Won't Fix", given the dependency on non-Wayland compositor + Gnome Shell).

So, if GNOME devs are not interested in fixing this for Wayland, that leaves [some other options][2].

[1]: https://github.com/Yubico/yubioath-flutter/issues/773#issuecomment-1603416435
[2]: https://github.com/Yubico/yubioath-flutter/issues/773#issuecomment-1603426008

liferooter:
Guys, just use portals for Wayland. It's generally impossible to implement a universal screenshot tool for Wayland without using something like portals, thanks to Wayland protocol. GNOME devs just can't physically fix that issue. Also screenshoters for Sway (that uses `wl-roots` extra protocols) won't work on GNOME or KDE, or generally any non-`wl-roots` Wayland compositor. Don't implement more and more broken hacks. Just use portals, generic API to do that.

Documentation for the portal you need: https://flatpak.github.io/xdg-desktop-portal/#gdbus-org.freedesktop.portal.Screenshot

jh-devv:
The same could be done for #1399

The current thing just uses an `zenity` based menu afaik but that probably needs to be addressed upstream https://github.com/mguelpruivo/flutter_file_picker/issues/1451

PR #1064, opened 2023-04-27
Add troubleshooting section + first solution

passbyval:
An issue I ran into while trying to import the icons on my M1 MacBoo pro. Took an hour or two to figure out the cause since Safari was silently unzipping my files.

### DIFF
--- a/doc/Custom_Account_Icons.adoc
+++ b/doc/Custom_Account_Icons.adoc
@@ -20,3 +20,20 @@ Once done, any account that has an issuer that is supported by the icon pack
 will display the custom icon instead of the standard colored circle with a
 letter. You can also use the `Custom icons` dialog to replace the icon pack, or
 to remove it.
+
+=== Troubleshooting
+
+==== Problem
+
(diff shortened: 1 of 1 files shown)


AdamVe:
Hi @passbyval, I am sorry to hear that you had struggles with the Custom Icon feature on your Mac and I appreciate that you took time to create this PR.

The instruction document was written to be future proof. If a user follows it step by step, they should be able to achieve the functionality. Because we cannot influence other systems, such as web browsers, operating systems, etc., we avoid mentioning them in our instructions. The proposed trouble-shooting section would make the document dependent on the functionality of the Safari Web browser and the instructions for trouble-shooting could become outdated without the team at Yubico even knowing.

Keeping the instructions written with focus on Yubico Authenticator makes the document accessible to users on all platforms. Please note that point 4 instructs the users to select an icon pack zip file.

Don’t hesitate to open new issues in the GitHub issue tracker, the team is always open to discuss the app’s improvements.

PR #1094, opened 2023-05-29
Fix right-click handling of OATH account

dainnilsson:
The popup menu doesn't reliably trigger on right-click, this fixes that.


AdamVe (approved):
Tested on Android and it worked correctly.

PR #1095, opened 2023-05-30
Management screen icon padding

AdamVe:
Updates gap between leading icon and title on Management screen.

Original:
management_orig

Fix:
image


dainnilsson:
I think this is better solved by changing the `horizontalTitleGap` which is explicitly set to 0 here. Just removing it completely uses the default value, but maybe that gap is too large?
image

dainnilsson (approved):

PR #1098, opened 2023-05-31
Android bite-size updates

AdamVe:
Contains following fixes:
- use correct flags for nfs adapter status change. This receiver must be exported to be able to receive system updates.
- in pages where we write text I unfocus text fields on submit. This fixes user experience on Android, when the virtual keyboard is shown again for a split second.
- changed log level of icon pack log messages from debug to traffic. They were too many.

### DIFF
--- a/android/app/src/main/kotlin/com/yubico/authenticator/MainActivity.kt
+++ b/android/app/src/main/kotlin/com/yubico/authenticator/MainActivity.kt
@@ -170,18 +170,17 @@ class MainActivity : FlutterFragmentActivity() {
     @SuppressLint("WrongConstant")
     override fun onStart() {
         super.onStart()
-        val receiverFlags = ContextCompat.RECEIVER_NOT_EXPORTED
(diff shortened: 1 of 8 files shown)


dainnilsson:
It feels like this is the wrong usage of the `TRAFFIC` level to me. I don't think the amount of data logged should determine which level is used, but the actual type of data logged. I do understand that the amount of logs makes it a bit unwieldy though, but maybe that is better solved with filtering.

AdamVe:
> It feels like this is the wrong usage of the `TRAFFIC` level to me. I don't think the amount of data logged should determine which level is used, but the actual type of data logged. I do understand that the amount of logs makes it a bit unwieldy though, but maybe that is better solved with filtering.

I understand what you mean, TRAFFIC is not correct for the icon pack log messages - it would be better with `VERBOSE` but we don't have this now. I will leave the logs on DEBUG level for now.

dainnilsson (approved):

PR #1101, opened 2023-06-09
[Android] Better user experience when using NFC  

AdamVe:
This PR iterates on the ways how we communicate that NFC is being used.

The goal is to make using of NFC YubiKeys the best possible experience for our users:

- long taking operations are visualized and users are aware that the application is reading the data
- ability to redo action over NFC when it failed, without the need to restarting the action from the UI
- beautiful living UI with animations and transitions (currently using Material UI bottom sheet widget)
- improved integration with Android - correct unfocusing of widgets so that the software keyboard does not popup unnecessarily. 

Current demo:

https://github.com/user-attachments/assets/5a382db8-2d84-4fcb-b9ed-971403e3c846


lib/app/views/main_page.dart
```diff
@@ -52,26 +52,40 @@ class MainPage extends ConsumerWidget {
...
+      if ((next.value != null) && (prev?.value != null)) {
+        // if there is change only in fipsApproved, don't pop anything
+        var nextInfo = next.value!.info;
+        var prevInfo = prev!.value!.info;
+
+        canPop =
+            prevInfo.copyWith(fipsApproved: nextInfo.fipsApproved) != nextInfo;
+      } else if (next.hasValue && (prev != null && prev.isLoading)) {
+        canPop = false;
+      }
```
dainnilsson, line 72: This is intended to pop when the device changes, so a better check would be to avoid popping when serial doesn't change:
```
final serial = next.value?.info.serial;
if (serial != null && serial == prev.value?.info.serial) {
  return;
}

Navigator.of(context).popUntil((route) {
  ...
```

AdamVe: thanks! this will avoid even more popping :)

PR #1108, opened 2023-06-18
add German translation

nebulon42:


lib/l10n/app_de.arb
```diff
@@ -0,0 +1,445 @@
...
+    "s_set_password": "Passwort setzen",
+    "s_password_set": "Passwort gesetzt",
+    "l_optional_password_protection": "Optionaler Passwortschutz",
+    "s_new_password": "Neues Passwort",
+    "s_current_password": "Aktuelles Passwort",
+    "s_confirm_password": "Passwort bestätigen",
+    "s_wrong_password": "Falsches Passwort",
+    "s_remove_password": "Passwort entfernen",
+    "s_password_removed": "Passwort entfernt",
+    "s_remember_password": "Remember password",
```
fdennis, line 220: This has not been translated.

lib/l10n/app_de.arb
```diff
@@ -0,0 +1,445 @@
...
+    "app_name": "Yubico Authenticator",
+
+    "s_save": "Speichern",
+    "s_cancel": "Abbrechen",
+    "s_close": "Schließen",
+    "s_delete": "Löschen",
+    "s_quit": "Beenden",
+    "s_unlock": "Entsperren",
+    "s_calculate": "Berechnen",
+    "s_label": "Label",
```
fdennis, line 32: On line 342 you use "die Beschriftung" for label. Could that be used here as well, instead of Label?

fdennis:
Thank you for the translation! I have added a few questions.

lib/l10n/app_de.arb
nebulon42: Oops, forgot that.

lib/l10n/app_de.arb
```diff
@@ -0,0 +1,445 @@
...
+    "s_reset_fido": "FIDO zurücksetzen",
+    "l_fido_app_reset": "FIDO Anwendung zurückgesetzt",
+    "l_press_reset_to_begin": "Drücken Sie Zurücksetzen um zu beginnen\u2026",
+    "l_reset_failed": "Fehler beim Zurücksetzen: {message}",
+    "@l_reset_failed" : {
+        "placeholders": {
+            "message": {}
+        }
+    },
+    "p_warning_factory_reset": "Achtung! Das löscht alle OATH TOTP/HOTP Konten unwiderbringlich von Ihrem YubiKey.",
```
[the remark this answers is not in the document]
nebulon42: Yes, it should be the latter.

nebulon42:
Thanks for the thorough review. I have updated the file according to your suggestions.

fdennis (approved):

AdamVe:
Thank you for your contribution @nebulon42, your work will be part of the next release 💯

PR #1143, opened 2023-08-07
Update Polish translations

Daviteusz:
I used Weblate for translation.
I tried various translation tools, but each of them generates a large number of "@".
Manually translating the file in a text editor to make it look identical to the original takes extra hours. :D 
I don't know how to do it better.


AdamVe:
@Daviteusz thank you so much for your help with the Polish localization. I have cleaned up the file with a script, so that those empty `"@_key": {}` are not bloating the file.

Daviteusz:
@AdamVe I tested reformat_strings.py and it works very well! I'll wait until PR is integrated and push out some new translation corrections 😃

AdamVe:
great :) currently that script is in testing phase, what it does is it cleans up all empty metadata strings and sorts the rest in alphabetical order with some extra rules.

PR #1144, opened 2023-08-08
Add SafeArea to FsDialog.

dainnilsson:
Fixes the new fullscreen dialog on Android where elements would show up behind system controls.


AdamVe (approved):
Looks great! Tested on Pixel 6.

PR #1148, opened 2023-08-15
Various PIV UI updates

dainnilsson:
- Certificate details in table, with copyable values.
- Better error handling for PIN management.


fdennis:
When prompting for mgmt key, do we not want to have a "Use default" checkbox?

dainnilsson:
> When prompting for mgmt key, do we not want to have a "Use default" checkbox?

For any key with metadata support (5.3 and newer) we auto-populate with the default management key iff it is in use, and make the field read-only. For older keys we could potentially add a button for it, so it's a bit dependent on how much effort we want to improve usage for older devices. I suppose <5.3 isn't that old, so it may make sense to be more helpful here.

fdennis:
If I write "UID=jsmith, DC=example" (note the whitespace) as subject when generating a cert nothing happens. The pop-up does not disappear.

dainnilsson:
> If I write "UID=jsmith, DC=example" (note the whitespace) as subject when generating a cert nothing happens. The pop-up does not disappear.

Sounds like a bug (not handling illegal values properly), but not really part of this PR.

lib/oath/views/account_dialog.dart
```diff
@@ -121,40 +121,51 @@ class AccountDialog extends ConsumerWidget {
...
+                        textAlign: TextAlign.center,
                       ),
+                      if (subtitle != null)
+                        Text(
+                          subtitle,
+                          softWrap: true,
+                          textAlign: TextAlign.center,
+                          // This is what ListTile uses for subtitle
+                          style:
+                              Theme.of(context).textTheme.bodyMedium!.copyWith(
```
AdamVe, line 159: can you use `Theme.of(context).listTileTheme.subtitleTextStyle?` to get the subtitle text style instead of hardcoding bodyMedium?

lib/piv/views/slot_dialog.dart
```diff
@@ -40,6 +61,32 @@ class SlotDialog extends ConsumerWidget {
       return const FsDialog(child: CircularProgressIndicator());
     }
 
+    TableRow detailRow(String title, String value) {
```
AdamVe, line 64: If possible rewrite this as a StatelessWidget subclass. Using functions which return Widget is an anti-pattern in flutter.

AdamVe (approved):

lib/piv/views/slot_dialog.dart
dainnilsson: Nope. We need a list of TableRow instances.

lib/oath/views/account_dialog.dart
dainnilsson: Unfotunately that would return null, since we don't define a ListTileTheme.

lib/piv/views/slot_dialog.dart
AdamVe: I see now 👍🏻

PR #1153, opened 2023-08-22
Add arb file reformatter

AdamVe:
Added `reformat_strings.py` which will sort an arb file by key name and remove empty property objects.

This is run on all arb files except app_en.arb during github actions builds and can generate diff errors.

Part of this PR is also reformatting of all arb files except app_en.arb.

The app_en.arb contains information which would be lost by reformatting and the file should be edited manually.

### DIFF
--- a/.github/workflows/check-strings.yml
+++ b/.github/workflows/check-strings.yml
@@ -15,6 +15,10 @@ jobs:
     - name: Ensure main locale is correct
       run: python check_strings.py lib/l10n/app_en.arb
 
+    - name: Reformat arb files
+      run: |
+        find lib/l10n/ -name "*.arb" -print0 | xargs -r -n 1 -0 ./reformat_strings.py
+
     - name: Check remaining locales
       run: |
         find lib/l10n/ -name "app_en.arb" -prune -o -name "*.arb" -print0 | xargs -r -n 1 -0 ./check_strings.py >> $GITHUB_STEP_SUMMARY || echo "::warning::Problems in locales"
--- a/.github/workflows/check-strings.yml
+++ b/.github/workflows/check-strings.yml
@@ -17,7 +17,7 @@ jobs:
 
     - name: Reformat arb files
       run: |
-        find lib/l10n/ -name "*.arb" -print0 | xargs -r -n 1 -0 ./reformat_strings.py
+        find lib/l10n/ -name "app_en.arb" -prune -o -name "*.arb" -print0 | xargs -r -n 1 -0 ./reformat_strings.py
 
     - name: Check remaining locales
       run: |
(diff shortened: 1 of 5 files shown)


AdamVe:
Closing this PR, will not be part of the repository. We will need to try a different way for easier maintaining of the localization files.

Daviteusz:
@AdamVe I managed to create a script that sorts strings identically to app_en.arb. Maybe it will be useful as a base.
```
# Function to change '@@locale' from 'en' to 'pl'
def change_locale(line):
    if '"@@locale": "en"' in line:
        line = line.replace('"@@locale": "en"', '"@@locale": "pl"')
    return line

# Opening app_en.arb as a source and unsorted translated arb_pl.arb file
with open('app_en.arb', 'r', encoding='utf-8') as en_file, open('app_pl.arb', 'r', encoding='utf-8') as pl_file:
    en_lines = en_file.readlines()
    pl_lines = pl_file.readlines()

# Function to return the key from the line
def get_key(line):
    parts = line.split(':', 1)
    return parts[0].strip() if len(parts) > 1 else None

# Function to increase indentation for placeholders
def increase_indentation(line, spaces=0):
    return ' ' * spaces + line

# Creating a dictionary for translations
translation_dict = {}
for pl_line in pl_lines:
    if pl_line.strip():
        key = get_key(pl_line)
        if key is not None:
            value = pl_line.split(':', 1)[1].strip()
            translation_dict[key] = value

# Opening the result file for writing
with open('app_plv2.arb', 'w', encoding='utf-8') as output_file:
    in_placeholders = False
    for en_line in en_lines:
        # change key'@@locale' from 'en' to 'pl'
        en_line = change_locale(en_line)
        key = get_key(en_line)
        if key is not None:
            if key == '"placeholders"':
                in_placeholders = True
                output_file.write(increase_indentation(en_line))
            else:
                if in_placeholders:
                    output_file.write(increase_indentation(en_line))
                elif key in translation_dict:
                    output_file.write(f'    {key}: {translation_dict[key]}\n')
                else:
                    output_file.write(en_line)
        else:
            # If a line does not contain a key, keep it
            output_file.write(en_line)
            if in_placeholders:
                in_placeholders = False

print("Plik 'app_plv2.arb' został wygenerowany z zwiększonym wcięciem wewnątrz 'placeholders' i zachowuje tłumaczenia z 'app_pl.arb'.")
```

AdamVe:
@Daviteusz Sorry for late reply, I saw your comment first now. Do you think we could work on this feature together? The idea to make the localization files (like your Polish file) to match the English "base" file is what we were thinking about, and I would like to adapt your script to suite our workflow.

Could you create a pull request to our main from your fork with the script from the comment? Then it will be easier to update and merge and keep you the contributor :)

Daviteusz:
@AdamVe  Check this out!
- #1243

PR #1196, opened 2023-09-18
Update macos release script

fdennis:


dainnilsson:
Pulled in changes from main and updated branch to no longer sign or use the arm64 helper, and fixed paths for PyInstaller 6.

PR #1209, opened 2023-09-27
Close Drawer when resized to expanded view.

dainnilsson:
This fixes an issue with closing the drawer after opening it, then resizing the window so it disappears and comes back.


AdamVe (approved):
LGTM

PR #1219, opened 2023-10-06
Use Python 3.12 for building the helper

dainnilsson:
Also updates Python dependencies, including PyInstaller which now results in a slightly different directory structure.


AdamVe (approved):
LGTM

PR #1234, opened 2023-10-22
Update Polish translation

Daviteusz:
- Fixed sorting to match app_en.arb
- Updated translations
- Removed unneeded whitespace characters


AdamVe:
Thank you @Daviteusz !

PR #1239, opened 2023-10-27
a11y: Minor fix(es)

fdennis:


lib/widgets/responsive_dialog.dart
```diff
@@ -53,14 +53,17 @@ class _ResponsiveDialogState extends State<ResponsiveDialog> {
...
           actions: widget.actions,
-          leading: IconButton(
-              icon: const Icon(Icons.close),
-              onPressed: widget.allowCancel
-                  ? () {
-                      widget.onCancel?.call();
-                      Navigator.of(context).pop();
-                    }
-                  : null),
+          leading: Semantics(
```
dainnilsson, line 56: Many of out other IconButtons have the tooltip field set. Does that also solve the issue wit screen readers? If yes, then I think that is a better approach a it also allows other users to see something more descriptive than just the icon.
We also have other instances of IconButton that do not provide tooltip (nor Semantics) that probably should in other places of the app.

fdennis: You're right! Fixed.

dainnilsson (approved):

PR #1243, opened 2023-11-02
Add arb_reformatter.py

Daviteusz:
Formats ARB files translated in various tools into ones that match the source file.

File translated in Weblate looks like:

After using the script:

Warning:
- Readme and Lint rules sections are copied from the source file
- Missing translations are filled from the source file.
- Code may look strange, but it works.

For my purposes it is already very good :)

### DIFF
--- a/arb_reformatter.py
+++ b/arb_reformatter.py
@@ -53,12 +53,32 @@ def update_arb_file(source_path, target_path, language_code):
         if ':' in line:
             key, value = line.split(':', 1)
             key = key.strip().strip('"')
-            translation_dict[key] = value.strip()
+            value = value.strip().strip(",").strip('"')
+            if not key.startswith("@"):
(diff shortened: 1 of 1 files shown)


AdamVe:
Hi, thanks again for your amazing contribution. I have updated your script with following:
- if a translation is missing, only add an empty line - this will allow us to see missing translations when running `flutter gen-l10n`
- make sure that we have correct `,` in the file - it might happen that when shuffling the file by lines, we get invalid json format
- treat `@_end` in a special way (in the en file it is last, so it will be placed last in the update file - in that case we need to remove any training comma)

AdamVe (approved):
Works nice!

AdamVe:
@Daviteusz I have merged the script now and will continue development in our repository. Again thank you for your contributions!

PR #1244, opened 2023-11-02
Use JSON parser in arb_reformatter.py

emlun:
Update the reformatting method to parse both dictionaries as JSON, recursively re-order the keys of the target to match that of the source, and then insert newlines where they appear in the source. Keys missing from the source are deleted from the target.


AdamVe (approved):
Looks good to me!

PR #1245, opened 2023-11-02
Make punctuation customizable per language in translation linter

emlun:
Japanese, for example, uses different punctuation than ASCII `.!?`.


AdamVe (approved):
LGTM!

PR #1254, opened 2023-11-07
Add plugin for parsing cmd line args

fdennis:


lib/desktop/init.dart
```diff
@@ -110,12 +111,20 @@ class _WindowEventListener extends WindowListener {
 }
 
 Future<Widget> initialize(List<String> argv) async {
-  _initLogging(argv);
+  var parser = ArgParser();
```
dainnilsson, line 114: `parser` and `args` below can be `final`.

lib/desktop/init.dart
```diff
@@ -266,29 +275,27 @@ Future<RpcSession> _initHelper(String exe) async {
...
-    String path;
+  String path;
+  if (args['log-file'] != null) {
     try {
-      path = argv[logFileIndex + 1];
+      path = args['log-file'];
     } catch (e) {
       throw InitializationException(
           'USAGE: Missing argument for option --log-file');
     }
```
dainnilsson, line 287: Does this still ever throw? Looks like we could remove the try-catch block now.

lib/desktop/init.dart
```diff
@@ -298,10 +305,9 @@ void _initLogging(List<String> argv) {
...
   });
 
-  final logLevelIndex = argv.indexOf('--log-level');
-  if (logLevelIndex != -1) {
+  if (args['log-level'] != null) {
     try {
-      final levelName = argv[logLevelIndex + 1];
+      final levelName = args['log-level'];
```
dainnilsson, line 310: Move declaration of `levelName` to the start, and do the null check on that, instead of repeated access of args['log-level']:

```
final levelName = args['log-level'];
if (levelName != null) {
  ...
}
```
Same for args['log-file'] above.

AdamVe (approved):
LGTM

dainnilsson (approved):

PR #1256, opened 2023-11-08
Reformat localization files

AdamVe:
Used `arb_reformatter.py` to update all existing arb files.

Updated the script to preserve specific characters for easier readability.


lib/l10n/app_de.arb
```diff
@@ -71,7 +88,7 @@
         }
     },
     "s_character_count": "Anzahl Zeichen",
-    "s_learn_more": "Mehr\u00a0erfahren",
+    "s_learn_more": "Mehr erfahren",
```
emlun, line 91: Note :point_up:

This is caused by the `ensure_ascii=False` argument in https://github.com/Yubico/yubioath-flutter/blob/2464733f71e5c6c8f188eb4a7eee0d46476be2d7/arb_reformatter.py#L58

in case we prefer keeping these the way they were.

AdamVe: good point, I will update and rather use the original `\u` format

emlun: I don't care either way, I just wanted to point this out so we're aware :smile:

AdamVe: Using `ensure_ascii=True` argument will convert all glyphs with code>128 to its `\u...` version which will make the files very hard to read. For example in German localization, instead of `"Ungültige Länge"` we will get `"Ung\u00fcltige L\u00e4nge"`, Japanese will be totally unreadable (`"パスキーが削除されました"` -> `"\u30d1\u30b9\u30ad\u30fc\u304c\u524a\u9664\u3055\u308c\u307e\u3057\u305f"`)

We need to use `ensure_ascii=False`

dainnilsson: Can we `ensure_ascii=False` and then manually replace certain characters with escape sequences? Eg. `target_text.replace(" ", r"\u00a0")`.

AdamVe: We could have a list of replacements for the sequences we use. I think currently it is the No-Break Space (U+00A0) and the Horizontal Ellipsis (U+2026).

I will update the script as part of this PR.

dainnilsson (approved):

arb_reformatter.py
```diff
@@ -18,6 +18,11 @@ def write_to_file(file_path, text):
     with open(file_path, 'w', encoding='utf-8') as file:
         file.write(text)
 
+# Translation table for unicode characters we want to keep readable.
```
emlun, line 21: Isn't it the opposite?

```suggestion
# Translation table for unicode characters we want to keep in escaped form.
```

AdamVe: yeah that's what I meant, I will use your suggestion 🥇 thanks!

AdamVe: Fixed!

emlun (approved):

PR #1283, opened 2023-11-30
[Android] Use only strings from ARB files 

AdamVe:
For unified localizations, this PR moves the strings defined in `strings.xml` to ARB and uses dart's l10n support to do the translations.

There are several other disadvantages:
- to get the strings we have to start flutter process, which is not needed when doing "just" ndef activity.
- the way how flutter localizations work makes us to call a specific method for each of the needed string

Advantages: 1 file with all localizable strings.


dainnilsson (approved):

AdamVe:
Superseded by #1291

PR #1312, opened 2023-12-20
Add localisation strings for FIDO error messages.

elibon99:


lib/fido/views/reset_dialog.dart
```diff
@@ -97,7 +97,15 @@ class _ResetDialogState extends ConsumerState<ResetDialog> {
                           final String errorMessage;
                           // TODO: Make this cleaner than importing desktop specific RpcError.
                           if (e is RpcError) {
-                            errorMessage = e.message;
+                            if (e.status == 'connection-error') {
+                              errorMessage = l10n.l_failed_connecting_to_fido;
+                            } else if (e.status == 'invalid-command') {
```
AdamVe, line 102: It smells to me that for 'invalid-command' we show that the inserted key is wrong. Is the 'invalid-command' sent from the helper only in this situation and if not, could we have more specific status, for example 'key-mismatch-error', to be used here instead?

helper/helper/fido.py
```diff
@@ -164,7 +172,11 @@ def reset(self, params, event, signal):
         self.ctap = Ctap2(connection)
         if target != _ctap_id(self.ctap):
             raise ValueError("Re-inserted YubiKey does not match initial device")
```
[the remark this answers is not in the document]
elibon99: I agree! I pushed a commit that implements a more specific exception class (`KeyMismatchException`) for this type of error.

lib/fido/views/reset_dialog.dart
elibon99: I pushed a commit that addresses this.

AdamVe (approved):
LGTM

PR #1313, opened 2023-12-20
Change FIDO fingerprint registration UI

elibon99:


AdamVe (approved):
LGTM, tested with different themes.

lib/fido/views/add_fingerprint_dialog.dart
```diff
@@ -193,35 +211,47 @@ class _AddFingerprintDialogState extends ConsumerState<AddFingerprintDialog>
                     },
                   ),
                 ),
-                LinearProgressIndicator(value: progress),
-                Padding(
-                  padding: const EdgeInsets.all(8.0),
-                  child: Text(_getMessage()),
-                ),
+                if (_fingerprint == null) ...[
```
fdennis, line 214: Is the `...[ ]` needed here?

elibon99: Good catch, you're right! I pushed a commit that removes the spread operator (`...`) at this and another place.

fdennis (approved):

PR #1325, opened 2024-01-05
Allow drag & drop in AccountList view

elibon99:
This PR enables the drag-and-drop functionality for OATH QR codes in the AccountList view. The overlay appearance during dragging has been improved for better visual feedback. See below for details.

Before:
drag_and_drop_before

After:
drag_and_drop_after

AccountList view:
drag_and_drop_account_list_view


lib/oath/views/oath_screen.dart
```diff
@@ -125,6 +131,27 @@ class _UnlockedViewState extends ConsumerState<_UnlockedView> {
     final numCreds = ref.watch(credentialListProvider(widget.devicePath)
         .select((value) => value?.length));
     final hasActions = ref.watch(featureProvider)(features.actions);
+    final qrScanner = ref.watch(qrScannerProvider);
+    final withContext = ref.read(withContextProvider);
+    final credentials = ref.read(credentialsProvider);
```
dainnilsson, line 136: I think these should be moved into the `onFileDropped` function below, to where they are used.

lib/oath/views/add_account_overlay.dart
```diff
@@ -0,0 +1,49 @@
...
+class AddAccountOverlay extends StatelessWidget {
+  const AddAccountOverlay({super.key});
+
+  @override
+  Widget build(BuildContext context) {
+    final l10n = AppLocalizations.of(context)!;
+
+    final theme = Theme.of(context);
+    final darkMode = theme.brightness == Brightness.dark;
+    // TODO: use color from theme.
```
dainnilsson, line 13: I think we should fix this before merging.

lib/oath/views/oath_screen.dart
elibon99: I pushed a commit that solves this.

elibon99:
I also pushed a commit that adds a generic widget (`FileDropOverlay`) for the file drop overlay. I believe it will come to good use in other parts of the application that require drag & drop.

lib/widgets/file_drop_overlay.dart
```diff
@@ -0,0 +1,53 @@
+import 'package:flutter/material.dart';
+
+class FileDropOverlay extends StatelessWidget {
+  final Widget? graphic;
+  final String? title;
+  final String? subTitle;
```
dainnilsson, line 6: Rename to `subtitle`, as in https://api.flutter.dev/flutter/material/ListTile/subtitle.html

dainnilsson (approved):

PR #1330, opened 2024-01-10
Fix parsing of `otpauth-migration` Uri.

elibon99:
This solves the issue regarding duplicate accounts not being recognized.

### DIFF
--- a/lib/oath/models.dart
+++ b/lib/oath/models.dart
@@ -188,22 +188,30 @@ class CredentialData with _$CredentialData {
 
     // Convert parsed credential values into CredentialData objects
     return splitCreds(base64.decode(uri.queryParameters['data']!))
-        .map((values) => CredentialData(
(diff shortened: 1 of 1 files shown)


lib/oath/models.dart
```diff
@@ -188,22 +188,30 @@ class CredentialData with _$CredentialData {
...
-            ))
-        .toList();
+        .map((values) {
+      String? issuer;
+      String name = utf8.decode(values[2], allowMalformed: true);
+      final nameIndex = name.indexOf(':');
+      if (nameIndex >= 0) {
+        issuer = name.substring(0, nameIndex);
+        name = name.substring(nameIndex + 1);
+      }
```
dainnilsson, line 198: We should only modify `name` if `value[3]` is null!

dainnilsson (changes requested):

dainnilsson (approved):

PR #1344, opened 2024-01-23
Key customizations

AdamVe:
Adds ability to customize key name and color for keys which can be distinguished by serial number.


lib/app/key_customization/key_customization.dart
```diff
@@ -0,0 +1,84 @@
...
+    final sha = _getSerialSha(serial);
+    _customizations[sha] = properties;
+  }
+
+  Future<void> write() async {
+    await _prefs.setString(_prefKeyCustomizations,
+        base64UrlEncode(utf8.encode(json.encode(_customizations))));
+  }
+
+  String _getSerialSha(String serialNumber) =>
```
dainnilsson, line 82: Why do we hash the serial number?

lib/app/key_customization/key_customization.dart
```diff
@@ -0,0 +1,84 @@
...
+      'display_color': customColor?.value.toRadixString(16),
+      'display_name': customName?.isNotEmpty == true ? customName : null
+    };
+    _log.debug('Setting key customization for $serial: $properties');
+    final sha = _getSerialSha(serial);
+    _customizations[sha] = properties;
+  }
+
+  Future<void> write() async {
+    await _prefs.setString(_prefKeyCustomizations,
```
dainnilsson, line 78: Why do we base64 encode the data?

lib/app/key_customization/models.dart
```diff
@@ -0,0 +1,46 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import 'dart:convert';
+import 'dart:ui';
+
+class KeyCustomization {
```
dainnilsson, line 20: Would it be better to use freezed for this data class? Instead of a dynamic `_properties` I would just add properties for the two values, and a custom JSON-serializer for the Color type.

lib/android/app_methods.dart
```diff
@@ -52,6 +52,10 @@ Future<int> getAndroidSdkVersion() async {
   return await appMethodsChannel.invokeMethod('getAndroidSdkVersion');
 }
 
+Future<int?> getPrimaryColor() async {
+  return await appMethodsChannel.invokeMethod('getPrimaryColor');
+}
```
dainnilsson, line 57: Might be nicer if this returns a Future<Color?> instead, by moving the conversion here.

lib/app/key_customization/views/key_customization_dialog.dart
```diff
@@ -0,0 +1,350 @@
...
+    }
+
+    final intValue = int.tryParse(hexValue, radix: 16);
+    if (intValue == null) {
+      return null;
+    }
+
+    return Color(intValue);
+  }
+}
```
dainnilsson, line 50: Why do we need this if it's unused?

lib/app/key_customization/models.dart
AdamVe: will look into this! thanks :)

lib/app/key_customization/key_customization.dart
AdamVe: You are right that hashing the serial number does not make too much sense in this context and I will remove it 👍🏻

lib/app/key_customization/key_customization.dart
AdamVe: Same here: does not make too much sense to base64 the json - it will make debugging harder as well. I will use the json string instead.

lib/app/key_customization/views/key_customization_dialog.dart
AdamVe: this has been moved to different file and is used now.

lib/app/views/device_picker.dart
```diff
@@ -197,53 +206,205 @@ class _DeviceRow extends StatelessWidget {
...
+        onSecondaryTapDown: isDesktop && menuItems.isNotEmpty
+            ? (details) {
+                showMenuFn(details);
+              }
+            : null,
+        onLongPressStart: isAndroid
+            ? (details) {
+                showMenuFn(details);
+              }
+            : null,
```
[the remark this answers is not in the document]
AdamVe: Great tip, will fix

lib/app/key_customization/key_customization.dart
AdamVe: fixed

lib/app/views/device_picker.dart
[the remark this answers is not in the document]
AdamVe: fixed

lib/android/app_methods.dart
AdamVe: fixed

lib/app/key_customization/key_customization.dart
```diff
@@ -41,24 +40,21 @@ class KeyCustomizationManager {
     }
 
     try {
-      final jsonString = String.fromCharCodes(base64Decode(pref));
-      return json.decode(utf8.decode(jsonString.codeUnits));
+      return json.decode(utf8.decode(pref.codeUnits));
```
dainnilsson, line 43: I think this can just be `json.decode(pref)`

lib/app/key_customization/key_customization.dart
```diff
@@ -70,15 +66,11 @@ class KeyCustomizationManager {
...
     _log.debug('Setting key customization for $serial: $properties');
-    final sha = _getSerialSha(serial);
-    _customizations[sha] = properties;
+    _customizations[serial] = properties;
   }
 
   Future<void> write() async {
     await _prefs.setString(_prefKeyCustomizations,
-        base64UrlEncode(utf8.encode(json.encode(_customizations))));
+        String.fromCharCodes(utf8.encode(json.encode(_customizations))));
```
dainnilsson, line 74: This can just be `json.encode(_customizations)`, no need to encode to UTF8 and then back again.

lib/app/key_customization/key_customization.dart
AdamVe: fixed!

lib/app/key_customization/key_customization.dart
AdamVe: fixed

dainnilsson (approved):

PR #1345, opened 2024-01-23
Fix rebuilding-loop of AppPages when not expanded.

dainnilsson:
Also fix deselect of FIDO items when resized to non-expanded.


AdamVe (approved):
LGTM

PR #1361, opened 2024-01-30
Hide the rail on phones when in portrait mode

dainnilsson:
We had this check before, but it was lost in the overhaul.


AdamVe (approved):
Tested and works as before, no navigation rail on my phone devices !

PR #1365, opened 2024-02-01
Add support for FIDO applet on Android

AdamVe:
With this change, users can use both OATH and FIDO functionality on keys which support those over NFC and USB.

Implements PIN management, passkey management and FIDO reset for the FIDO applet.

### DIFF
--- a/android/app/src/main/kotlin/com/yubico/authenticator/fido/FidoPinStore.kt
+++ b/android/app/src/main/kotlin/com/yubico/authenticator/fido/FidoPinStore.kt
@@ -1,0 +1,33 @@
+/*
+ * Copyright (C) 2024 Yubico.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.yubico.authenticator.fido
+
+class FidoPinStore {
+    private var pin: CharArray? = null
+
+    fun hasPin(): Boolean {
+        return pin != null
+    }
+
+    fun getPin(): CharArray {
+        return pin!!
+    }
+
+    fun setPin(newPin: CharArray?) {
+        pin = newPin?.clone()
+    }
+}
--- a/android/app/src/main/kotlin/com/yubico/authenticator/fido/data/Session.kt
+++ b/android/app/src/main/kotlin/com/yubico/authenticator/fido/data/Session.kt
@@ -1,3 +1,19 @@
+/*
+ * Copyright (C) 2024 Yubico.
+ *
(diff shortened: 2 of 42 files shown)


android/app/src/main/kotlin/com/yubico/authenticator/fido/FidoPinStore.kt
```diff
@@ -0,0 +1,33 @@
...
+    fun hasPin(): Boolean {
+        return pin != null
+    }
+
+    fun getPin(): CharArray {
+        return pin!!
+    }
+
+    fun setPin(newPin: CharArray?) {
+        pin = newPin?.clone()
```
dainnilsson, line 31: Should we zero the old pin here if it is set before re-assigning it?

android/app/src/main/kotlin/com/yubico/authenticator/fido/data/Session.kt
```diff
@@ -0,0 +1,116 @@
...
+)
+
+fun Map<String, Any?>.getBoolean(
+    key: String,
+    default: Boolean = false
+): Boolean = get(key) as? Boolean ?: default
+
+fun Map<String, Any?>.getOptionalBoolean(
+    key: String
+): Boolean? = get(key) as? Boolean
```
dainnilsson, line 40: Maybe make these extensions on InfoData instead?

android/app/src/main/kotlin/com/yubico/authenticator/fido/data/Session.kt
```diff
@@ -0,0 +1,116 @@
...
+                    bioEnroll = null,
+                    alwaysUv = false
+                ),
+                aaguid = ByteArray(0),
+                minPinLength = 0,
+                forcePinChange = false
+            ),
+            unlocked = false,
+            initialized = false
+        )
```
dainnilsson, line 113: I don't like the "initialized" variable on the Session objects, that's what `null` should be for. I understand that the problem with this may be that we need to differentiate on the Flutter side between "loading" and "waiting for tap", but I think we can solve that in a better way.

We use AsyncData here, which has a loading state (no arguments), an error state (with an error), and a data state with a (Fido|Oath)State. One solution could be to make it a type of error (eg. NotInitializedError) that can have special handling.

AdamVe: Good idea, I will update this to be some kind of error state 👍🏻

android/app/src/main/kotlin/com/yubico/authenticator/fido/data/Session.kt
AdamVe: Will investigate this, makes sense to avoid bloating standard types with extensions.

android/app/src/main/kotlin/com/yubico/authenticator/fido/FidoPinStore.kt
AdamVe: Will fix 👍🏻

android/app/src/main/kotlin/com/yubico/authenticator/fido/data/Session.kt
AdamVe: fixed

android/app/src/main/kotlin/com/yubico/authenticator/fido/data/Session.kt
AdamVe: Fixed!

dainnilsson (approved):

PR #1366, opened 2024-02-02
Key generation progressbar

dainnilsson:
Adds progress bar to key generation dialog, since key generation can be slow. Also fixes ResponsiveDialog to respect `allowCancel` when in dialog mode.


AdamVe (approved):
LGTM

PR #1383, opened 2024-02-08
Improve PIV public key handling

dainnilsson:
Displays more information about the key in a PIV slot (requires YK 5.3), and allows export of the public key when no certificate is loaded in the slot. Also adds public key as output format when generating new keys.


AdamVe (approved):
LGTM

PR #1384, opened 2024-02-08
Hide labels on segment buttons when they don't fit

dainnilsson:
Also adds icons, and tooltips for when labels are not show.


jokkon:
👍

AdamVe (approved):

PR #1386, opened 2024-02-09
Adds keys to reset ButtonSegments

AdamVe:
The keys are needed for integration tests to click on the correct ButtonSegment. 
As ButtonSegment has no `key` property, it is added on the icon which is always shown. Tested on Mac.


jokkon:
Verified keys not missing and doing their work. 👍

PR #1395, opened 2024-02-13
Consistent padding for `footnote`

elibon99:
This make sure we have consistent padding between `footnote` and remaining content.

### DIFF
--- a/lib/app/views/app_page.dart
+++ b/lib/app/views/app_page.dart
@@ -202,7 +202,8 @@ class AppPage extends StatelessWidget {
           Align(
             alignment: centered ? Alignment.center : Alignment.centerLeft,
             child: Padding(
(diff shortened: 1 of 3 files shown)


lib/app/views/app_page.dart
```diff
@@ -197,11 +197,13 @@ class AppPage extends StatelessWidget {
             child: _buildTitle(context),
           ),
         builder(context, expanded),
-        if (actionsBuilder != null)
+        if (actionsBuilder != null &&
+            actionsBuilder!(context, expanded).isNotEmpty)
```
dainnilsson, line 201: Instead of invoking actionsBuilder twice, make a variable further up:
`final actions = actionsBuilder?.call(context, expanded) ?? [];`

elibon99: I agree! Pushed a commit that resolves this.

dainnilsson (approved):

PR #1411, opened 2024-02-20
Use official Python docker image for helper build

dainnilsson:
Instead of using an ubuntu base image, use one from Python with the correct version of Python already installed.


jokkon:
Build works well on Ubuntu 20.04 and 22.04. 👍

PR #1449, opened 2024-03-19
Add command to delete key

elibon99:


AdamVe (approved):
I am approving this change and have added some thoughts in the comments, which you are welcome to consider. I tested the branch on MacOS and it works well as it is. 

It would be also nice to evolve the PIV integration tests and add scenarios validating the new functionality.

dainnilsson (approved):

PR #1457, opened 2024-03-21
Add link to yubicloud otp upload

elibon99:


lib/l10n/app_en.arb
```diff
@@ -622,6 +622,7 @@
     "s_no_export": "No export",
     "s_export": "Export",
     "l_export_configuration_file": "Export configuration to file",
+    "l_export_can_be_added_to": "Exported credential can be added to",
```
fdennis, line 625: Perhaps "Exported credentials"?

elibon99: I agree! Pushed a commit that changes this

fdennis (approved):

lib/l10n/app_en.arb
elibon99: @emlun Good point! I have created another PR (#1458) that tries to resolve this.

PR #1458, opened 2024-03-22
Interpolate yubiotp upload link in translation string

elibon99:


lib/otp/views/configure_yubiotp_dialog.dart
```diff
@@ -390,6 +384,28 @@ class _ConfigureYubiOtpDialogState
...
+                : TextSpan(text: e),
+          )
+        ],
+      ),
+    );
+  }
+
   TextSpan _createUploadOtpLink(BuildContext context) {
     final theme = Theme.of(context);
     final uploadOtpUri = Uri.parse('https://upload.yubico.com');
```
emlun, line 411: Looks like this is now duplicated above?

lib/otp/views/configure_yubiotp_dialog.dart
```diff
@@ -390,6 +384,28 @@ class _ConfigureYubiOtpDialogState
     );
   }
 
+  RichText _createUploadText(BuildContext context, AppLocalizations l10n) {
+    final uploadText = l10n.l_exported_can_be_uploaded_at(uploadOtpUri.host);
+    final host = uploadOtpUri.host;
+    final parts = uploadText.split(RegExp('(?=$host)|(?<=$host)'));
+
+    return RichText(
```
emlun, line 392: Am I reading this right that this is splitting the translated string to find where the URL is, and wrapping the URL with a clickable link element (because rich component can't be directly interpolated in the translation, I guess?)?

If so, is this just splitting _before_ the URL, or _around_ the URL? For example, what happens if you set the translation string to `"Exported credential can be uploaded at {url}, if that's something you would like to do."`?

elibon99: > Am I reading this right that this is splitting the translated string to find where the URL is, and wrapping the URL with a clickable link element (because rich component can't be directly interpolated in the translation, I guess?)?

Correct! 

> If so, is this just splitting before the URL, or around the URL? For example, what happens if you set the translation string to "Exported credential can be uploaded at {url}, if that's something you would like to do."?

Splitting occurs _around_ the URL since we are alternating postive and negative look around. The delimiter (the URL in this case) will be included in `parts`. So your example works. I tested it and the value of `parts` is `["Exported credential can be uploaded at "," upload.yubico.com", ," if that's something you would like to do."]`. 
Note that there is leading whitespace on the last two elements. This is automatically removed in `TextSpan`.

lib/otp/views/configure_yubiotp_dialog.dart
elibon99: Good catch! Will remove that.

lib/otp/views/configure_yubiotp_dialog.dart
emlun: Oh, I see! I didn't notice that only the `host` part is supplied as the placeholder, so I was thinking this should clash with the URL scheme and path, if any. But it makes a lot of sense to only display the host part even if there is a path component in the actual URL. Nice!

emlun (approved):
:+1:

PR #1467, opened 2024-03-25
Android: add support for FIDO fingerprints

AdamVe:


lib/android/fido/state.dart
```diff
@@ -165,26 +166,152 @@ final androidFingerprintProvider = AsyncNotifierProvider.autoDispose
...
+    return Completer<List<Fingerprint>>().future;
   }
 
   @override
   Stream<FingerprintEvent> registerFingerprint({String? name}) {
     final controller = StreamController<FingerprintEvent>();
+    const registerEvents = EventChannel('android.fido.registerFp');
+
+    final registerFpSub =
+        registerEvents.receiveBroadcastStream().skip(1).listen((event) {
```
dainnilsson, line 197: What is the event we're skipping here?

lib/android/fido/state.dart
```diff
@@ -165,26 +166,152 @@ final androidFingerprintProvider = AsyncNotifierProvider.autoDispose
...
+              .add(FingerprintEvent.complete(Fingerprint.fromJson(resultJson)));
+        } else {
+          // TODO abstract platform errors
+          final errorStatus = resultJson['status'];
+          if (errorStatus != 'user-cancelled') {
+            throw RpcError(errorStatus, 'Platform error: $errorStatus', {});
+          }
+        }
+
+        await controller.sink.close();
```
dainnilsson, line 244: Should we close in a `finally`-clause to avoid leaving the steam open if there is an exception?

dainnilsson (changes requested):

android/app/src/main/kotlin/com/yubico/authenticator/fido/FidoManager.kt
```diff
@@ -413,6 +441,152 @@ class FidoManager(
...
+        fidoSession: YubiKitFidoSession,
+        clientPin: ClientPin,
+        pinUvAuthToken: ByteArray
+    ): List<FidoFingerprint> {
+        val bioEnrollment =
+            FingerprintBioEnrollment(fidoSession, clientPin.pinUvAuth, pinUvAuthToken)
+
+        val enrollments: Map<ByteArray, String?> = bioEnrollment.enumerateEnrollments()
+        return enrollments.map { enrollment ->
+            FidoFingerprint(Base64.toUrlSafeString(enrollment.key), enrollment.value)
```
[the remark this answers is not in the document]
AdamVe: I updated this (see https://github.com/Yubico/yubioath-flutter/pull/1467/commits/6fc18f8fddae3404b41a42584ffeb23fc429641f) and tested on Android x Desktop. The strings for unnamed fingerprints are rendered the same way.

lib/android/fido/state.dart
AdamVe: fixed

lib/android/fido/state.dart
AdamVe: We decided to refactor this later, the backing live data is not the best way to solve the fingerprint registration, but using skip(1) works.

dainnilsson (approved):
Nice! And I guess we can just use toHexString/hexToByteArray once they are stable?

AdamVe:
> Nice! And I guess we can just use toHexString/hexToByteArray once they are stable?

Yes, and if they had been removed (the risk exists for ExperimentalStdlibApi) then we would provide our own implementations.

PR #1489, opened 2024-04-03
Uncustomizable devices will use last used color

AdamVe:
Before this change, uncustomizable devices used the default color.

### DIFF
--- a/lib/app/state.dart
+++ b/lib/app/state.dart
@@ -152,7 +152,6 @@ final primaryColorProvider = Provider<Color>((ref) {
   final data = ref.watch(currentDeviceDataProvider).valueOrNull;
   final defaultColor = ref.watch(defaultColorProvider);
   if (data != null) {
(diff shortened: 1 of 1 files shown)


dainnilsson:
Code-wise this looks fine. The changes to the comments are a bit misleading, as they imply any change has to come from device connect/disconnect. I'd go with something along the lines of: `A device is selected, use its custom color, or the default color` and `No custom color, use the last used color, or the default color`

AdamVe:
> Code-wise this looks fine. The changes to the comments are a bit misleading, as they imply any change has to come from device connect/disconnect. I'd go with something along the lines of: `A device is selected, use its custom color, or the default color` and `No custom color, use the last used color, or the default color`

I agree, the way I wrote it is not clear, as the change can happen on switching the keys (not only connect/disconnect). The code is now clear enough to follow so we don't need the comments.

dainnilsson (approved):

PR #1491, opened 2024-04-04
Display FIDO PIN Block Status

elibon99:


lib/fido/views/pin_entry_form.dart
```diff
@@ -113,12 +114,14 @@ class _PinEntryFormState extends ConsumerState<PinEntryForm> {
               autofillHints: const [AutofillHints.password],
               controller: _pinController,
               focusNode: _pinFocus,
-              enabled: !_blocked && (_retries ?? 1) > 0,
+              enabled: !widget._state.authBlocked &&
```
AdamVe, line 117: use the `authBlocked` variable instead of `widget._state.authBlocked`

AdamVe (commented):
See my comments

lib/l10n/app_en.arb
```diff
@@ -295,6 +295,8 @@
     "l_fido_pin_protection_optional": "Optional FIDO PIN protection",
     "l_enter_fido2_pin": "Enter the FIDO2 PIN for your YubiKey",
     "l_pin_blocked_reset": "PIN is blocked; factory reset the FIDO application",
+    "l_fido_pin_blocked": "PIN is blocked",
+    "l_note_attempts_remaining": "Note: {retries} attempt(s) remaining",
```
AdamVe, line 299: Flutter has support for Plurals in localized strings: https://docs.flutter.dev/ui/accessibility-and-internationalization/internationalization#placeholders-plurals-and-selects

dainnilsson:
`"l_fido_pin_blocked": "PIN is blocked"` - Can we remove `fido` from the key, and have this as a general PIN string? Nothing in the value seems FIDO specific.

`"l_note_attempts_remaining": "Note: {retries} attempt(s) remaining",` - Do we need a new string for this when we already have `l_attempts_remaining`?

lib/l10n/app_en.arb
elibon99: Sounds interesting! There's definitely a lot of strings we could use that for. I will investigate this and hopefully incorporate it in a future PR.

dainnilsson (approved):

PR #1493, opened 2024-04-05
Ensure initial currentSection is `Section.Home` for Android

elibon99:


AdamVe (approved):
Tested with various YubiKeys: the app starts in Home section and tapping/connecting a YubiKey will switch to the preferred section of that key (Accounts on Neo, Passkeys on SKY etc).

PR #1525, opened 2024-04-16
[Android] Always request focus for OATH and FIDO2 auth

AdamVe:
When switching sections on Android, the text Password (OATH) and PIN (FIDO2) text fields did not get focus automatically, this explicitly requests the focus when the widgets are initialized.


braathen (approved):
it works from a UX perspective!

PR #1530, opened 2024-04-17
Fix typos

AdamVe:


integration_test/testdoc.adoc
```diff
@@ -15,7 +15,7 @@ the flutter test framework.
 
 To run the tests you need to specify the serial number of a Yubikey which the tests are able to run
 on (OBS this key will be reset by running the tests, so don't use one of your personal keys). This
-either be defined as an anvironment variable on execution, or in a CSV-file with serial numbers:
+either be defined as an environment variable on execution, or in a CSV-file with serial numbers:
```
fdennis, line 18: Perhaps also add "should" here:
`This should either be defined...`

fdennis (approved):

integration_test/testdoc.adoc
AdamVe: I updated the paragraph as well.

PR #1554, opened 2024-05-13
Flexible layout for Passkeys and OATH

elibon99:
This update adds support for changing the layout presentation of OATH accounts and Passkeys.


lib/oath/views/oath_screen.dart
```diff
@@ -53,6 +53,19 @@ import 'key_actions.dart';
...
+        OathLayout.list => Symbols.list,
+        OathLayout.grid => Symbols.grid_view,
+        OathLayout.mixed => Symbols.vertical_split
+      };
+  String getDisplayName(AppLocalizations l10n) => switch (this) {
+        OathLayout.list => l10n.s_list_layout,
+        OathLayout.grid => l10n.s_grid_layout,
+        OathLayout.mixed => l10n.s_mixed_layout
+      };
+}
```
dainnilsson, line 67: Should these maybe be members of OathLayout instead? Or are they specific to this view only (eg. other views would use different values)?

elibon99: If we were to use `_icon` and `getDisplayName` in any other OATH view, the values would be the same. However, I think this is the only OATH view that will be using these values which is why I implemented the members here (and because they are UI specific).

PR #1593, opened 2024-07-03
Bump Android dependencies

AdamVe:
Significant updates:
- uses Kotlin 2.0.0
- libraries and build libraries updated to latest versions
- fixed low hanging lint fruits


android/app/build.gradle
```diff
@@ -94,23 +94,16 @@ dependencies {
...
 
     implementation 'com.github.tony19:logback-android:3.0.0'
 
-    implementation('commons-codec:commons-codec') {
-        version {
-            // use version 1.15 for compatibility reasons
-            strictly '1.15'
-        }
-    }
-
```
AdamVe, line 113: This is not needed anymore as yubikit-android 2.6.0 uses own implementation of base32 and does not depend on commons-codec. 
We used commons-coded in yubikit-android only for b32 codec, however version 1.16 was not compatible with older Android APIs (and that is why we forced 1.15 in YA), so it was decided to implement our own b32 and not use the library at all. See https://github.com/Yubico/yubikit-android/pull/126

dainnilsson (approved):

PR #1598, opened 2024-07-10
Dain/fips status

dainnilsson:


android/app/src/main/kotlin/com/yubico/authenticator/device/Info.kt
```diff
@@ -57,6 +57,8 @@ data class Info(
     val fipsCapable: Int,
     @SerialName("fips_approved")
     val fipsApproved: Int,
+    @SerialName("reset_blocked")
```
AdamVe, line 60: If possible, could we make the order of the properties same? In the dart model file, `resetBlocked` is before the `fipsCapable`. As this data class is serialized to json, the order does not matter for functionality, but it looks nicer if one sees the same order.

dainnilsson: Sure, fine by me!

PR #1617, opened 2024-08-06
Enable Enterprise Attestation

elibon99:


android/app/src/main/kotlin/com/yubico/authenticator/fido/FidoActionDescription.kt
```diff
@@ -26,6 +26,7 @@ enum class FidoActionDescription(private val value: Int) {
     DeleteFingerprint(4),
     RenameFingerprint(5),
     RegisterFingerprint(6),
+    EnableEnterpriseAttestation(7),
     ActionFailure(7);
```
dainnilsson, line 30: 7 is used twice

android/app/src/main/kotlin/com/yubico/authenticator/fido/FidoManager.kt
```diff
@@ -603,6 +611,44 @@ class FidoManager(
...
+            val uvAuthProtocol = getPreferredPinUvAuthProtocol(fidoSession.cachedInfo)
+            val clientPin = ClientPin(fidoSession, uvAuthProtocol)
+            val token = if (pinStore.hasPin()) {
+                clientPin.getPinToken(
+                    pinStore.getPin(),
+                    getPinPermissionsACFG(fidoSession),
+                    null
+                )
+            } else null
+            val config = Config(fidoSession, uvAuthProtocol, token)
```
dainnilsson, line 626: This line will fail if Config.isSupported() returns false, throwing an exception. Since this can only work when it is supported, the `getPinPermissionsACFG` method can just be replaced with `ClientPin.PIN_PERMISSON_ACFG`, can't it?

dainnilsson (changes requested):

android/app/src/main/kotlin/com/yubico/authenticator/fido/FidoManager.kt
AdamVe: yes, I am updating this

android/app/src/main/kotlin/com/yubico/authenticator/fido/FidoActionDescription.kt
AdamVe: Fixed

android/app/src/main/kotlin/com/yubico/authenticator/fido/FidoManager.kt
AdamVe: Rewrote this part.

android/app/src/main/kotlin/com/yubico/authenticator/fido/FidoManager.kt
```diff
@@ -603,6 +606,47 @@ class FidoManager(
...
+                }
+            } else {
+                logger.debug("authenticatorConfig not supported, ignoring call to enableEnterpriseAttestation")
+            }
+
+            return@useSession JSONObject(
+                mapOf(
+                    "success" to true,
+                )
+            ).toString()
```
dainnilsson, line 648: It seems strange to me that we would silently ignore and return `success: true` when the YubiKey doesn't support Entrprise Attestation. Shouldn't that result in a failure?

AdamVe: sure! I will return a failure when this method is called and the authenticator does not support authenticatorConfig.

AdamVe: updated

dainnilsson (approved):

PR #1626, opened 2024-08-13
Adds FIPS status indicators

dainnilsson:


lib/piv/views/piv_screen.dart
```diff
@@ -57,6 +58,14 @@ class _PivScreenState extends ConsumerState<PivScreen> {
   Widget build(BuildContext context) {
     final l10n = AppLocalizations.of(context)!;
     final hasFeature = ref.watch(featureProvider);
+    final (fipsCapable, fipsApproved) = ref
+            .watch(currentDeviceDataProvider)
+            .valueOrNull
+            ?.info
+            .getFipsStatus(Capability.piv) ??
+        (false, false);
```
AdamVe, line 66: this could be a turned into provider on its own and reused in several places making the code shorter

lib/piv/views/actions.dart
```diff
@@ -310,7 +310,15 @@ class PivActions extends ConsumerWidget {
 }
 
 List<ActionItem> buildSlotActions(
-    PivState pivState, PivSlot slot, AppLocalizations l10n) {
+    PivState pivState, PivSlot slot, bool fipsUnready, AppLocalizations l10n) {
+  if (fipsUnready) {
+    return [
+      ActionItem(
+          icon: const Icon(Symbols.add),
+          title: 'Provision slot',
```
AdamVe, line 318: replace with a localization or add TODO

lib/home/views/home_screen.dart
```diff
@@ -195,6 +256,29 @@ class _DeviceContent extends ConsumerWidget {
                 .titleSmall
                 ?.apply(color: Theme.of(context).colorScheme.onSurfaceVariant),
           ),
+        if (deviceData.info.pinComplexity)
+          Padding(
+            padding: const EdgeInsets.only(top: 12),
+            child: RichText(
```
AdamVe, line 262: why not use just a wrap or row?

dainnilsson: Verical alignment between the icon and text is more "correct" when done this way.

AdamVe (approved):
LGTM

lib/piv/views/piv_screen.dart
dainnilsson: I think it's a balance between repeating some code, and having too many providers with very low usage. If this pattern is repeated too much it makes sense to make it a Provider of its own, but as it is now I don't really think it's warranted.

PR #1642, opened 2024-08-21
Avoid having both 'Configure YK' and 'Expand sidebar' icons at the same time

fdennis:


lib/app/views/app_page.dart
```diff
@@ -684,8 +684,7 @@ class _AppPageState extends ConsumerState<AppPage> {
                 ),
           actions: [
             if (widget.actionButtonBuilder == null &&
-                (widget.keyActionsBuilder != null &&
-                    (!hasManage || !showDetailView)))
+                (widget.keyActionsBuilder != null && (!hasManage)))
```
elibon99, line 687: Maybe remove the nested parentheses around `!hasManage`  to improve readability?

fdennis: Agreed, fix coming up!

elibon99 (approved):

PR #1643, opened 2024-08-21
Fixes to OATH screen and actions

AdamVe:
This change fixes following:
- add account button visible even if the key is not FIPS approved
- adds badges and icons to actions, if a FIPS capable is not setup correctly (missing password)
- on Android, the device info is updated after reset and password change so that Flutter gets correct values for FIPS directly after the user action.

This was tested with FIPS capable keys and non-FIPS keys (so that nothing was broken) over USB and NFC connecitons.

### DIFF
--- a/lib/oath/views/utils.dart
+++ b/lib/oath/views/utils.dart
@@ -180,7 +180,7 @@ Future<void> addOathAccount(BuildContext context, WidgetRef ref,
   }
 }
 
-Future<void> setManagePassword(BuildContext context, WidgetRef ref,
+Future<void> managePassword(BuildContext context, WidgetRef ref,
     DevicePath devicePath, OathState oathState) async {
   await showBlurDialog(
     context: context,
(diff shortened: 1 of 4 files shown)


lib/oath/views/utils.dart
```diff
@@ -178,3 +179,11 @@ Future<void> addOathAccount(BuildContext context, WidgetRef ref,
     );
   }
 }
+
+Future<void> setManagePassword(BuildContext context, WidgetRef ref,
```
dainnilsson, line 183: I guess this name is from "Set or Manage Password", but it kind of looks like a setter. I think we should just name it `managePassword`, which matches the name of the dialog as well.

dainnilsson (approved):

lib/oath/views/utils.dart
AdamVe: will fix! 👍🏻

PR #1661, opened 2024-08-29
Show explanation page when using phone without SCP11b support

AdamVe:


lib/l10n/app_en.arb
```diff
@@ -236,6 +236,8 @@
     "s_unsupported_yk": "Unsupported YubiKey",
     "s_yk_not_recognized": "Device not recognized",
     "p_operation_failed_try_again": "The operation failed, please try again.",
+    "l_scp11b_nfc_technology_missing": "Configuration not supported",
```
dainnilsson, line 239: I think we should rename this key to `l_configuration_unsupported` or something. Since the string is pretty generic, we might be able to reuse it in other locations. The next one I would name `p_scp_unsupported` as we use `unsupported` in other places (we can also use `not_supported` in these, as we have that in other places too), and it's more general than SCP11b. Usually we add the `_desc` suffix if it's clearly a description that relates to another string, but if we change the title to be more generic we shouldn't need it here.

AdamVe: updated

dainnilsson (approved):

PR #1677, opened 2024-09-06
Update l10n

AdamVe:
Adds new strings and fixes rename account error toast error.

### DIFF
--- a/lib/l10n/app_en.arb
+++ b/lib/l10n/app_en.arb
@@ -446,6 +446,12 @@
     "s_rename_account": "Rename account",
     "l_rename_account_desc": "Edit the issuer/name of the account",
     "s_account_renamed": "Account renamed",
+    "l_rename_account_failed": "Failed renaming account: {message}",
+    "@l_rename_account_failed": {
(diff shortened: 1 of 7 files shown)


lib/l10n/app_en.arb
```diff
@@ -915,6 +921,13 @@
     "s_nfc_dialog_fido_rename_fingerprint": "Action: rename fingerprint",
     "s_nfc_dialog_fido_failure": "FIDO operation failed",
 
+    "@_nfc": {},
+    "s_nfc_ready_to_scan": "Ready to scan",
+    "s_nfc_scanning": "Scanning\u2026",
+    "s_nfc_tap_your_yubikey": "Tap your YubiKey",
+    "l_nfc_failed_to_scan": "Failed to scan, try again",
+    "s_nfc_done": "Done",
```
dainnilsson, line 929: This could be just `s_done` and could then be reused in other places.

AdamVe: fixing!

dainnilsson (approved):

PR #1734, opened 2024-11-25
Add icon for F-Droid 

AdamVe:
Adds an icon file which will be used by F-Droid.

The used directory structure is [Triple-T](https://f-droid.org/en/docs/All_About_Descriptions_Graphics_and_Screenshots/#triple-t-structure).


fastlane/metadata/android/en-US/images/icon.png
AdamVe, line 1: This file does not render correctly in GitHub, but it is valid "link" to existing file.

dainnilsson (approved):

PR #1745, opened 2024-12-09
Attempt to repair/restore a broken preferences file

fdennis:
See also https://github.com/flutter/flutter/issues/89211

### DIFF
--- a/lib/desktop/init.dart
+++ b/lib/desktop/init.dart
@@ -131,16 +131,18 @@ Future<Widget> initialize(List<String> argv) async {
   try {
     prefs = await SharedPreferences.getInstance();
   } catch (error) {
-    // Attempt to repair the broken preferences file
     Directory appSupportDirectory = await getApplicationSupportDirectory();
     String appDataPath =
         path.join(appSupportDirectory.path, 'shared_preferences.json');
+    _log.warning(
+        'Failed to load the preferences file at $appDataPath. Attempting to repair it.');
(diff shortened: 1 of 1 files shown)


lib/desktop/init.dart
```diff
@@ -125,7 +127,28 @@ Future<Widget> initialize(List<String> argv) async {
   _initLogging(args);
 
   await windowManager.ensureInitialized();
-  final prefs = await SharedPreferences.getInstance();
+  SharedPreferences prefs;
+  try {
+    prefs = await SharedPreferences.getInstance();
+  } catch (error) {
+    // Attempt to repair the broken preferences file
+    Directory? appSupportDirectory = await getApplicationSupportDirectory();
```
elibon99, line 135: Should this be nullable? Looks like `getApplicationSupportDirectory` always returns a `Directory` (creates on if it doesn't exist).

fdennis: Good catch! I think we can fix that and also remove the `doesDirectoryExist ` check, since `getApplicationSupportDirectory()` will create the directory if it does not exist.

elibon99 (approved):
LGTM!

dainnilsson (changes requested):
Tested on Linux, both repair and delete works as expected. I would like some logging to be added. Failure to load the file should probably be a `warning`, as should failure to repair, informing on deletion.

dainnilsson (approved):

PR #1751, opened 2025-01-07
roll back to python 3.12

PhilBrk8:
The virtual environment was set to 3.13 already, which broke the build-helper.sh install due to pyinstaller.


dainnilsson:
Thanks! We've instead bumped the dependencies so that everything builds on Python 3.13. Please let us know if you're having any trouble building from `main`.

PR #1755, opened 2025-01-09
Remove `AppTextFormField`

elibon99:
AppTextFormField is mainly meant to be used in a Form widget. However, we never use it that way which is the reason for deleting it. We can achieve the same functionality with AppTextField.


AdamVe (approved):
LGTM

PR #1783, opened 2025-02-06
Add language selector and move custom icons to settings

elibon99:


AdamVe (approved):
LGTM! Tested on Android devices and found some glitches which I fix in near future.

PR #1795, opened 2025-02-20
Enable ICU Message Syntax on a few strings for pluralization support

braathen:


AdamVe (approved):
LGTM, next step is to add space before `other` and use `1` instead of `{retries}` in the `=1` variants.

PR #1822, opened 2025-03-11
Toggle Applications

AdamVe:
Adds implementation for Toggle applications on Android.

- Users can from the home page select "Toggle Application" action which will show a view with enabled and disabled Applications (or Interfaces on NEO) which the user can change.
- Works for both NFC and USB connections.
- The PR adds several workaround for YubiKey NEO so that it is possible to read get correct USB interfaces over NFC connection
- Adds logic for selecting which Section is shown when a YubiKey is connected -> the last section or the most appropriate section.
- Adds USB overlay which is shown when the YubiKey enters the reclaim period
- Forbids changing the mode to OTP only


android/app/src/main/kotlin/com/yubico/authenticator/MainViewModel.kt
```diff
@@ -32,20 +37,54 @@ enum class OperationContext(val value: Int) {
...
+                (if (isNfc) config.enabledCapabilities.nfc else config.enabledCapabilities.usb)
+                    ?: 0
+
+            capabilitiesToContext.forEach { entry ->
+                if (capabilities and entry.key.bit == entry.key.bit) {
+                    operationContexts.add(entry.value)
+                }
+            }
+
+            if (name.contains("bio", ignoreCase = true) && !isNfc) {
```
AdamVe, line 64: use FormFactor here

android/app/src/main/kotlin/com/yubico/authenticator/Util.kt
```diff
@@ -18,4 +18,7 @@ package com.yubico.authenticator
 
 fun ByteArray.asString() = joinToString(
     separator = ""
-) { b -> "%02x".format(b) }
\ No newline at end of file
+) { b -> "%02x".format(b) }
+
+fun AppContextManager.supportsAny(contexts: List<OperationContext>) =
```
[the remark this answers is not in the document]
AdamVe: does Kotlin support this directly with `any` ?

android/app/src/main/kotlin/com/yubico/authenticator/device/DeviceManager.kt
```diff
@@ -161,6 +119,7 @@ class DeviceManager(
                 listener.onConnected(yubiKeyDevice)
             }
         }
+        clearDeviceInfoOnDisconnect = true
```
AdamVe, line 122: review

dainnilsson (approved):

PR #1826, opened 2025-03-13
Disable cancel during FIDO credential deletion

elibon99:


AdamVe (approved):
LGTM, tested on MacOS

PR #1827, opened 2025-03-14
Fix broken systray icon on Linux

dainnilsson:


AdamVe (approved):
Tested on Ubuntu 24.10 and the systray has a drop down and shows pinned accounts correctly (once the app window has been shown).

PR #1838, opened 2025-03-18
Programmatically open info popup menu

elibon99:
Instead of using PopupMenuButton when `!displayDialog` we can use the same component and open the menu/dialog programmtically. This avoids problems with widgets being unmounted while the menu/dialog is opened.


AdamVe (approved):
LGTM

PR #1866, opened 2025-04-28
Fix shortcuts and focus-related issues

elibon99:


AdamVe (approved):
LGTM, tested keyboard navigation on desktop.

AdamVe (approved):

PR #1881, opened 2025-05-15
Jokkon/linux arm64 2

jokkon:
Build pipeline for building Yubico Authenticator for Linux arm64 platform. Build has been tested and verified on Raspberry pi 5. This will not be a supported platform yet, however that is the long-term goal.

### DIFF
--- a/.github/workflows/linux-arm64.yml
+++ b/.github/workflows/linux-arm64.yml
@@ -1,0 +1,127 @@
+name: Linux
+
+on: [push, pull_request]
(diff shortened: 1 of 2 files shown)


dainnilsson:
According to https://github.blog/changelog/2025-01-16-linux-arm64-hosted-runners-now-available-for-free-in-public-repositories-public-preview/ it looks like we should be able to switch to `ubuntu-22.04-arm` for this!

PR #1893, opened 2025-05-20
Fix the tray icon when running in a flatpak/snap sandbox

sumwale:
Pass only the icon name rather than the path when running in a flatpak/snap sandbox since the path will not be visible in the host environment.

The upstream merged PR https://github.com/leanflutter/tray_manager/pull/43 already has the change that passes through the name without the path when running in a sandbox.

### DIFF
--- a/lib/desktop/systray.dart
+++ b/lib/desktop/systray.dart
@@ -22,6 +22,7 @@ import 'package:flutter/foundation.dart';
 import 'package:flutter_riverpod/flutter_riverpod.dart';
 import 'package:local_notifier/local_notifier.dart';
 import 'package:logging/logging.dart';
+import 'package:tray_manager/src/helpers/sandbox.dart';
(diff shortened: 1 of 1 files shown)


AdamVe:
@sumwale Thanks for the fix, I will merge your work 👍🏻 

I will have to do some updates because of the following warning:

`Import of a library in the 'lib/src' directory of another package • lib/desktop/systray.dart:25:8 • implementation_imports`

sumwale:
@AdamVe Thanks. I tested manually and missed running the test target. The tray_manager package has a pending PR https://github.com/leanflutter/tray_manager/pull/78 to enhance the `runningInSandbox()` check for docker/podman/... containers, so wanted to avoid copying the code of the method. I suppose it should be exposed as a public API in the tray_manager package for proper usage, so will propose a PR there.

AdamVe:
@sumwale that would be great, once that is a public API we will be happy to update the check.

PR #1931, opened 2025-06-23
Refactor UI tests

dainnilsson:
Provides a new foundation to build robust tests on.


integration_test/otp_test.dart
```diff
@@ -1,165 +1,240 @@
...
+        );
+        expect(
+          $.read(otpStateProvider(data.node.path)).value?.slot1Configured,
+          isFalse,
+        );
+
+        if (state.slot1Configured) {
+          // Slot 1 was pre-programmed, swap back
+          await $.read(otpStateProvider(data.node.path).notifier).swapSlots();
+        } else {
```
elibon99, line 60: Since we perform the swap above, Slot 1 will always be empty, so deleting it will fail. We should probably either delete Slot 2 here or always perform the swap back and only delete Slot 1 if it was not pre-programmed.

PR #1935, opened 2025-06-30
Re-factor settings view

elibon99:


fdennis:
The string "Translations are community translated on Crowdin." is not visible when the menu bar on the right is open. Perhaps it could be a good idea to have it there as well?

AdamVe (approved):
LGTM. tested on Mac and Android.

PR #1948, opened 2025-08-07
Block X25519 if output format is not public key

elibon99:


AdamVe (approved):
Tested on mac, cannot reproduce the issue.

PR #1979, opened 2025-09-05
Change preferred connection type based on device

dainnilsson:


elibon99 (approved):
LGTM! Tested on MacOS

PR #1988, opened 2025-09-09
[Android] set default log level on app startup

AdamVe:
This PR prevents debug and trace log messages from appearing in release builds before the configured minimum log level (INFO) is applied.

Previously, during early startup the logging system had not yet set the effective level, so initial log events were emitted at DEBUG/TRACE and reached Logcat. We now initialize logging at the earliest safe point and set a conservative default (INFO) so no low-level messages leak.

### DIFF
--- a/android/app/src/main/kotlin/com/yubico/authenticator/MainActivity.kt
+++ b/android/app/src/main/kotlin/com/yubico/authenticator/MainActivity.kt
@@ -807,5 +807,5 @@ class MainActivity : FlutterFragmentActivity() {
     private fun isPortraitOnly() = resources.getBoolean(R.bool.portrait_only)
 
     private val defaultLogLevel =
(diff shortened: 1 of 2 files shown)


android/app/src/main/kotlin/com/yubico/authenticator/MainActivity.kt
```diff
@@ -801,4 +805,7 @@ class MainActivity : FlutterFragmentActivity() {
     }
 
     private fun isPortraitOnly() = resources.getBoolean(R.bool.portrait_only)
+
+    private val defaultLogLevel =
+        if (BuildConfig.DEBUG) Log.LogLevel.DEBUG else Log.LogLevel.INFO
```
dainnilsson, line 810: This was already happening before, right? Is that set in some other place, and can that now be removed?

PR #2131, opened 2026-03-31
Read out text fields

oskardanielsson-yubico:
Makes VoiceOver read out the text field. However this fix may not be optimal as it's not the most intuitive.
Issues:


fdennis:
This is a bug in Flutter. See https://github.com/flutter/flutter/issues/120779

PR #2133, opened 2026-03-31
Add labels/tooltips to elements

oskardanielsson-yubico:
adds labels and improves navigation


lib/app/views/app_page.dart
```diff
@@ -858,7 +858,7 @@ class _AppPageState extends ConsumerState<AppPage> {
                       Symbols.dock_to_left,
                       fill: showExpandedSideMenuBar ? 1 : 0,
                       weight: 600.0,
-                      semanticLabel: sideMenuBarText,
+                      semanticLabel: l10n.s_toggle_menu_bar,
```
elibon99: This removes `sideMenuBarText` that shifts between "Collapse menu bar" and "Expand menu bar". I think this should be kept for accessibility reasons, since it announces the collapsed and expanded states.

lib/widgets/visibility_toggle_button.dart
```diff
@@ -0,0 +1,58 @@
+/*
+ * Copyright (C) 2024-2026 Yubico.
```
elibon99: Should probably be just "Copyright (C) 2026 Yubico"

elibon99 (approved):

elibon99 (approved):

fdennis (approved):

PR #2141, opened 2026-04-09
Announce character counter updates for screen readers

fdennis:


AdamVe (approved):
LGTM, all text fields with max chars covered, tested on Android.

PR #2150, opened 2026-04-13
a11y: Announce radio buttons as radio buttons correctly

oskardanielsson-yubico:


lib/app/views/settings_page.dart
```diff
@@ -493,75 +493,102 @@ class _LanguageView extends ConsumerWidget {
...
+                : null,
+            title: Transform.translate(
+              offset: Offset(isDesktop ? 4 : 0, 0),
+              child: Text(e.getNativeDisplayName()),
+            ),
+            value: e,
+            toggleable: true,
+          ),
+        ),
+        if (infoButton != null)
```
elibon99, line 587: This puts the info button outside of the `RadioListTile`. Is it possible to add the info button inside the `RadioListTile` with `secondary` and still ensure the tiles are announced correctly?

PR #2177, opened 2026-04-30
fix(android): handle NFC HAL crash on startup

### LINKED ISSUE #2167: Android: Yubico Authenticator crashing on Launch when NFC is activated
**Application Version:** Yubico Authenticator v7.3.2 on Android (from Google Play Store)

**OS Version:** Android 16 (LineageOS 23.2)

**Issue**: When I turn on NFC and then open Yubico Authenticator, the app immediatly crashes turning off NFC in the Process. I can consistently reproduce the issue.

**Workaround:** Open Yubico Authenticator with NFC disabled. Then turn on NFC.

Does anyone have the same issue or is this specific to my device/OS configuration?

Please let me know if you need additional information.

AdamVe:
## Summary

Prevents app crash when the NFC HAL service dies during startup (observed on OnePlus 7 Pro with LineageOS). Adds retry logic to recover NFC discovery after transient failures.

## Changes

- Catch RuntimeException from NfcAdapter.enableReaderMode when the NFC service is dead (DeadObjectException)
- Retry NFC discovery up to 3 times with 2-second delay between attempts, allowing the HAL to restart
- Set hasNfc = false on failure to prevent stopNfcDiscovery from accessing uninitialized YubiKit state
- Convert startNfcDiscovery from expression body to block body to support recursive retry calls

Fixes #2167

### DIFF
--- a/android/app/src/main/kotlin/com/yubico/authenticator/MainActivity.kt
+++ b/android/app/src/main/kotlin/com/yubico/authenticator/MainActivity.kt
@@ -82,6 +82,7 @@ import java.security.Security
 import java.util.concurrent.Executors
 import javax.crypto.Mac
 import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.Job
 import kotlinx.coroutines.delay
 import kotlinx.coroutines.launch
 import kotlinx.coroutines.withContext
@@ -100,6 +101,7 @@ class MainActivity : FlutterFragmentActivity() {
     private val nfcConfiguration = NfcConfiguration().timeout(5000)
 
     private var hasNfc: Boolean = false
+    private var nfcRetryJob: Job? = null
 
     private lateinit var yubikit: YubiKitManager
 
@@ -170,11 +172,11 @@ class MainActivity : FlutterFragmentActivity() {
             hasNfc = true
         } catch (_: NfcNotAvailable) {
             hasNfc = false
-        } catch (e: Exception) {
-            logger.error("Start NFC discovery failed (attempt {}): {}", retryCount + 1, e.message)
+        } catch (e: RuntimeException) {
+            logger.error("Start NFC discovery failed (attempt {})", retryCount + 1, e)
             hasNfc = false
             if (retryCount < NFC_DISCOVERY_MAX_RETRIES) {
-                lifecycleScope.launch(Dispatchers.Main) {
+                nfcRetryJob = lifecycleScope.launch(Dispatchers.Main) {
                     delay(NFC_DISCOVERY_RETRY_DELAY_MS)
                     startNfcDiscovery(retryCount + 1)
                 }
@@ -188,6 +190,8 @@ class MainActivity : FlutterFragmentActivity() {
     }
 
     private fun stopNfcDiscovery() {
+        nfcRetryJob?.cancel()
+        nfcRetryJob = null
         if (hasNfc) {
             yubikit.stopNfcDiscovery(this)
             logger.debug("Stopped nfc discovery")
--- a/android/app/src/main/kotlin/com/yubico/authenticator/MainActivity.kt
+++ b/android/app/src/main/kotlin/com/yubico/authenticator/MainActivity.kt
@@ -82,6 +82,7 @@ import java.security.Security
 import java.util.concurrent.Executors
 import javax.crypto.Mac
 import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.delay
 import kotlinx.coroutines.launch
 import kotlinx.coroutines.withContext
 import org.bouncycastle.jce.provider.BouncyCastleProvider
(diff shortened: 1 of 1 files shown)


android/app/src/main/kotlin/com/yubico/authenticator/MainActivity.kt
```diff
@@ -154,20 +155,36 @@ class MainActivity : FlutterFragmentActivity() {
...
-        hasNfc = false
+            hasNfc = true
+        } catch (_: NfcNotAvailable) {
+            hasNfc = false
+        } catch (e: Exception) {
+            logger.error("Start NFC discovery failed (attempt {}): {}", retryCount + 1, e.message)
+            hasNfc = false
+            if (retryCount < NFC_DISCOVERY_MAX_RETRIES) {
+                lifecycleScope.launch(Dispatchers.Main) {
+                    delay(NFC_DISCOVERY_RETRY_DELAY_MS)
```
emlun, line 180: This idea might not make sense in this context, but have you considered starting with a shorter delay and increasing the delay with each attempt? For example starting with 100 ms and doubling on each failure (that would make 7 attempts within 6.3 s) or starting with 250 ms and adding +250 ms on each failure (also 7 attempts within 5.25 s). I guess it depends on how likely it is that a short delay is "usually" enough, vs. if the failure cause usually takes around a few seconds to resolve.

android/app/src/main/kotlin/com/yubico/authenticator/MainActivity.kt
```diff
@@ -154,20 +155,36 @@ class MainActivity : FlutterFragmentActivity() {
...
-        hasNfc = true
-    } catch (_: NfcNotAvailable) {
-        hasNfc = false
+            hasNfc = true
+        } catch (_: NfcNotAvailable) {
+            hasNfc = false
+        } catch (e: Exception) {
+            logger.error("Start NFC discovery failed (attempt {}): {}", retryCount + 1, e.message)
+            hasNfc = false
+            if (retryCount < NFC_DISCOVERY_MAX_RETRIES) {
```
emlun, line 178: Nitpicking, but: the name says "retries", but I think the logic actually makes this "attempts" - `NFC_DISCOVERY_MAX_RETRIES = 3` will make the code abort after 3 failures, not 4 (1 initial + 3 retries). 🙂 So the total time the retry loop will spend is `(NFC_DISCOVERY_MAX_RETRIES - 1) * NFC_DISCOVERY_RETRY_DELAY_MS = 4000 ms`, not `MAX_RETRIES * RETRY_DELAY_MS = 6000 ms` as one might assume from just reading the definition of those two.

android/app/src/main/kotlin/com/yubico/authenticator/MainActivity.kt
AdamVe: The crash log shows the NFC HAL takes at least 1.5s to attempt recovery (and still hasn't fully initialized by the end of the log), so 2s is a conservative starting point. Since I can't reproduce this and don't know the actual recovery time, I'd rather start with a simple fixed delay and confirm it works with the reporter before optimizing the retry strategy. We can always refine to exponential backoff later if needed.

android/app/src/main/kotlin/com/yubico/authenticator/MainActivity.kt
AdamVe: Good catch on the naming, but I think it's actually correct as "retries": with `MAX_RETRIES = 3`, the initial call has `retryCount = 0`, and `retryCount < 3` schedules retries at `retryCount` 1, 2, and 3; so 1 initial + 3 retries = 4 total attempts (which is also why the giving-up log uses `MAX_RETRIES + 1`), with 3 × 2000 ms = 6000 ms of delay between them. Let me know if I'm misreading it!

emlun: Ah yes, sorry, my mistake. It'll indeed be `if (2 < 3) { startNfcDiscovery(3)` for a total of 4 invocations. What I said would be true if that condition was checked at the beginning of the function rather than before the retry call, as is the case here.

android/app/src/main/kotlin/com/yubico/authenticator/MainActivity.kt
```diff
@@ -154,20 +155,36 @@ class MainActivity : FlutterFragmentActivity() {
...
             }
-        }
 
-        hasNfc = true
-    } catch (_: NfcNotAvailable) {
-        hasNfc = false
+            hasNfc = true
+        } catch (_: NfcNotAvailable) {
+            hasNfc = false
+        } catch (e: Exception) {
```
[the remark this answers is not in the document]
AdamVe: Fixed in 949a9a2

android/app/src/main/kotlin/com/yubico/authenticator/MainActivity.kt
```diff
@@ -154,20 +155,36 @@ class MainActivity : FlutterFragmentActivity() {
...
+            logger.error("Start NFC discovery failed (attempt {}): {}", retryCount + 1, e.message)
+            hasNfc = false
+            if (retryCount < NFC_DISCOVERY_MAX_RETRIES) {
+                lifecycleScope.launch(Dispatchers.Main) {
+                    delay(NFC_DISCOVERY_RETRY_DELAY_MS)
+                    startNfcDiscovery(retryCount + 1)
+                }
```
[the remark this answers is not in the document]
AdamVe: Fixed in 949a9a2

android/app/src/main/kotlin/com/yubico/authenticator/MainActivity.kt
```diff
@@ -154,20 +155,36 @@ class MainActivity : FlutterFragmentActivity() {
...
-        }
 
-        hasNfc = true
-    } catch (_: NfcNotAvailable) {
-        hasNfc = false
+            hasNfc = true
+        } catch (_: NfcNotAvailable) {
+            hasNfc = false
+        } catch (e: Exception) {
+            logger.error("Start NFC discovery failed (attempt {}): {}", retryCount + 1, e.message)
```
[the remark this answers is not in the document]
AdamVe: Fixed in 949a9a2

Issue #2178, opened 2026-05-01
Android: NFC tap opens browser to my.yubico.com instead of launching app


thomasbuilds:
## Summary

Since commits 9edf9070 ("add ACTION_VIEW", 2025-04-23) and 44e62168 ("add autoVerify", 2025-04-24) I think in AndroidManifest.xml, tapping a YubiKey via NFC no longer launches the app. Instead, Android displays a prompt offering to open the link `https://my.yubico.com/yk/…` in a browser, and clicking it opens the browser on that URL.

## Steps to reproduce (I use grapheneos)

1. Install Yubico Authenticator from the Play Store on Android 16 or later.
2. Ensure "Launch Authenticator on NFC tap" is enabled in app settings
3. Close / background the app.
4. Tap a YubiKey to the NFC reader.
5. **Expected:** app opens and communicates with the key.
6. **Actual:** Android shows a prompt "Open link my.yubico.com?" / "Open with browser", and tapping it opens the browser.

## Possible Root cause

The `AliasNdefActivity` activity-alias in AndroidManifest.xml now declares **two** intent-filters for the same host (`my.yubico.com`):

```xml
<!-- Added April 2025 — turns this component into an Android App Link handler -->
<intent-filter android:autoVerify="true">
    <action android:name="android.intent.action.VIEW" />
    <data android:host="my.yubico.com" android:scheme="https" />
    <category android:name="android.intent.category.DEFAULT" />
    <category android:name="android.intent.category.BROWSABLE" />
</intent-filter>

<!-- Pre-existing NFC dispatch entry point -->
<intent-filter>
    <action android:name="android.nfc.action.NDEF_DISCOVERED" />
    <category android:name="android.intent.category.DEFAULT" />
    <data android:host="my.yubico.com" android:scheme="https" />
</intent-filter>
```

Adding `android:autoVerify="true"` triggers **Digital Asset Links** verification against `https://my.yubico.com/.well-known/assetlinks.json`. That JSON lists Yubico's production signing certificate. When verification fails or is pending, Android 12+ strict Safer Intents handling no longer auto-selects the unverified handler. NFC tag dispatch then falls through past `ACTION_NDEF_DISCOVERED` and resolves the URI via `ACTION_VIEW` / browser instead.

[...]

PR #2188, opened 2026-05-07
fix(android): split NDEF and App Link aliases

AdamVe:
## Summary

Splits `AliasNdefActivity` so that `NDEF_DISCOVERED` and the `ACTION_VIEW` + `autoVerify` App Link filter live on separate `activity-alias` entries. This is a structural cleanup that decouples NFC tag dispatch from App Link verification state.

This change is unlikely to be a user-visible fix on its own. The reported regression on Android 16 is rooted in a platform behavior change: from Android 16, URL tags are dispatched via standard `ACTION_VIEW` and only verified App Link handlers receive them - `NDEF_DISCOVERED` is no longer tried for URL tags regardless of how the manifest is structured. For builds whose App Link is not verified (sideloads, builds signed with non-Play keys, distributions on systems where DAL verification doesn't run), the user-visible workaround is to associate the domain manually one time via Settings → Apps → Yubico Authenticator → Open by default → add `my.yubico.com`.

The split may still help on Android 12–15 with unverified App Links, where the reporter's "Safer Intents poisoning" theory could apply (combined alias treated as fully unverified, blocking NDEF dispatch as well). That path could not be tested locally.

## Changes

- Split `AliasNdefActivity` into two aliases targeting the same `NdefActivity`:
  - `AliasNdefActivity` keeps only `android.nfc.action.NDEF_DISCOVERED` (Android 15 and below, plus the fallback path for builds whose App Links are not verified).
  - New `AliasNdefAppLinkActivity` carries `android.intent.action.VIEW` + `android:autoVerify="true"` + `BROWSABLE` (required by Android 16+  for NFC URL dispatch).
- `ActivityUtil.enableAppNfcDiscovery()` / `disableAppNfcDiscovery()` now toggle both aliases together so the "Launch on NFC tap" preference applies consistently across Android versions.
- Kept `COMPONENT_ENABLED_STATE_DISABLED` for the off path - `DEFAULT` would leave the aliases visible in the system NFC chooser when the preference is off.

Refs #2178

### DIFF
--- a/android/app/src/main/AndroidManifest.xml
+++ b/android/app/src/main/AndroidManifest.xml
@@ -83,6 +83,8 @@
             android:theme="@style/NdefActivityTheme">
         </activity>
 
+        <!-- NFC tag dispatch on Android 15 and below, and as a fallback for
+             builds whose App Links are not verified. -->
         <activity-alias
             android:name=".AliasNdefActivity"
             android:exported="true"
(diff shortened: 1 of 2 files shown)


AdamVe:
Closing without merging.

Testing on stock Android 16+ (locally) and GrapheneOS (via the issue reporter) confirmed that this PR does not change user-visible behaviour.

The reported symptom is caused by the Android 16 platform behaviour change for NFC URL dispatch, and the actual fix for affected installs is a one-time manual domain association in system settings (**Settings → Apps → Yubico Authenticator → Open by default → add `my.yubico.com`**), not a manifest change.

The alias split is structurally cleaner and theoretically removes a "Safer Intents poisoning" failure mode on Android 12–15 with unverified App Links, but that path could not be reproduced or verified, so there's no measurable benefit to merging.

See #2178 for the workaround and full discussion

PR #2190, opened 2026-05-07
android: prompt for my.yubico.com link association

### LINKED ISSUE #2189: Show domain association prompt automatically when NFC launch is enabled on Android 16
On Android 16, NFC taps on YubiKey no longer trigger `NDEF_DISCOVERED` but instead route through `ACTION_VIEW`. For the app to handle these taps, `my.yubico.com` must be manually associated in Settings → Apps → Yubico Authenticator → Open by default. Most users will never find this on their own.

**Possible suggested fix:**

Use `DomainVerificationManager` to check whether `my.yubico.com` is already associated when:
- The app is launched for the first time after install/update, or
- The user enables "Launch Authenticator on NFC tap" in settings

If the domain is not yet associated, show a brief explanation and trigger the system dialog so the user can approve it in one tap instead of digging through system settings manually.

thomasbuilds:
Attemps to fix #2189 

### Background

On Android 16 (API 36), NFC taps on a YubiKey no longer fire `NDEF_DISCOVERED`. The tag's NDEF record is treated as a regular `https://my.yubico.com/...` URL and dispatched via `ACTION_VIEW`. The app's manifest already declares an `autoVerify` intent filter for that host, but unless my.yubico.com is associated with the app (via Digital Asset Links auto-verification, or the user enabling it under **Settings → Apps → Yubico Authenticator → Open by default**), `ACTION_VIEW` falls through to a browser and the launch-on-tap setting silently does nothing.

Android 12–15 are unaffected: `NDEF_DISCOVERED` is still delivered there regardless of domain verification state.

### Approach

Use `DomainVerificationManager` (API 31+) to query whether `my.yubico.com` is associated with the app, and show a dialog with a direct link to the system "Open by default" settings page when it isn't.

The status check covers both conditions that allow `ACTION_VIEW` to be routed to the app:

1. `hostToStateMap[my.yubico.com]` must be `DOMAIN_STATE_VERIFIED` or `DOMAIN_STATE_SELECTED`.
2. `userState.isLinkHandlingAllowed` must be true — if the user disabled "Open supported links" globally for the app, `ACTION_VIEW` won't be routed here even when the host is individually verified or selected.

If either condition fails the prompt is shown.

**Note on "one tap" UX:** Android has no public API to programmatically grant per-domain link approval. `Settings.ACTION_APP_OPEN_BY_DEFAULT_SETTINGS` is the deepest available deep link — it lands directly on the per-app "Open by default" page where the user can enable "Open supported links" and verify `my.yubico.com` in two taps rather than navigating through system settings manually.

The prompt is shown at two points:

- **When the user selects a launch tap action** (`NfcTapAction.launch` / `NfcTapAction.launchAndCopy`) in NFC settings.
- **At startup**, once per app build (tracked in `SharedPreferences`), when a launch action is already configured. Users are reminded after install/update but not on every launch.

### Changes

**Kotlin / Android**

- New `DomainVerificationHelper` — guards on API 36, checks `isLinkHandlingAllowed` before `hostToStateMap`, swallows runtime exceptions to behave gracefully on quirky OEM builds.
- `MainActivity.AppMethodChannel` exposes `getDomainVerificationStatus` and `openDomainVerificationSettings` on the existing `app.methods` channel.

**Dart / Flutter**

- app_methods.dart — typed `DomainVerificationStatus` enum + channel bindings.
- domain_association_prompt.dart — `promptDomainAssociation()` (settings change) and `maybePromptDomainAssociationOnStartup()` (startup). Uses the shared `BasicDialog` widget.
- init.dart — invokes the startup variant in a post-frame callback when a launch action is configured.
- settings_page.dart — invokes the on-change variant when the user selects `launch` or `launchAndCopy`.

**Localization**

- app_en.arb — adds `s_link_my_yubico` and `p_link_my_yubico_desc`. Reuses existing `s_open_settings`. check_strings.py passes.

### Compatibility

- `minSdk = 24`, `compileSdk = 37`. The prompt is entirely suppressed below API 36 (`STATUS_UNSUPPORTED`). `DomainVerificationManager` calls are guarded on API 31 as required by the SDK, but the outer API-36 gate makes them unreachable on older releases.
- The legacy `NDEF_DISCOVERED` intent filter and `AliasNdefActivity` are unchanged — Android ≤15 behavior is unaffected.
- No new permissions.

### DIFF
--- a/android/app/src/main/kotlin/com/yubico/authenticator/DomainVerificationHelper.kt
+++ b/android/app/src/main/kotlin/com/yubico/authenticator/DomainVerificationHelper.kt
@@ -1,0 +1,128 @@
+/*
+ * Copyright (C) 2026 Yubico.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.yubico.authenticator
+
+import android.content.Context
+import android.content.Intent
+import android.content.pm.verify.domain.DomainVerificationManager
+import android.content.pm.verify.domain.DomainVerificationUserState
+import android.net.Uri
+import android.os.Build
+import android.provider.Settings
+import org.slf4j.LoggerFactory
+
+/**
+ * Helper for inspecting and managing the system-level "Open by default" status
+ * of the my.yubico.com domain.
+ *
+ * On Android 16 (API 36), NFC taps on a YubiKey no longer broadcast
+ * NDEF_DISCOVERED but launch the my.yubico.com URL via ACTION_VIEW. The app is
+ * registered to handle that URL but, unless Digital Asset Links auto-verifies
+ * the association or the user manually opted in via Settings → Apps → Yubico
+ * Authenticator → Open by default, ACTION_VIEW falls back to a browser. This
+ * helper lets the app detect that situation and deep-link the user to the
+ * relevant settings page.
+ *
+ * The DomainVerificationManager API is available from Android 12 (API 31), but
+ * the underlying problem only exists on Android 16+: on earlier releases NFC
+ * taps still arrive via NDEF_DISCOVERED regardless of link verification state.
+ * Therefore [getStatus] returns [STATUS_UNSUPPORTED] below API 36 to avoid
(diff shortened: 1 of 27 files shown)


AdamVe:
Thanks for the contribution, @thomasbuilds! I like the idea a lot. I'm on other tasks right now so review may take a while, but I'll get to it as soon as I can.

AdamVe:
Please run `pre-commit run --all-files` to identify the pre-commit issues. Usually this will update formatting of files.

thomasbuilds:
ubuntu apt repos have been down for nearly 1 week so I can't even install the package!!
~~will try to fix it manually~~

AdamVe:
No problem! When I start working with the PR I will make sure the precommit passes. Thanks for trying!

thomasbuilds:
I ran the pre-commit

thomasbuilds:
Can you verify that also on regular Android phones the deep link isn't enabled automatically when installing app on playstore?
If it is, then this is only a GrapheneOS issue and this PR should be closed?

PR #2208, opened 2026-06-03
fix(qrscanner): register observer after init

AdamVe:
## Summary

`StateChangeObserver` was never registered because `controller.cameraInfo`
is `null` immediately after `bindToLifecycle()` — camera binding happens
asynchronously. The observer is now attached inside
`initializationFuture.addListener`, guaranteeing `cameraInfo` is non-null
when the subscription is made.

## Changes

- Wrap `cameraState` observer registration in
  `controller.initializationFuture.addListener` so it runs after the
  camera is fully initialized


teddy-chavez (approved):
good stuff!

PR #2233, opened 2026-07-13
Release 7.4.1

AdamVe:


lib/app/views/navigation.dart
```diff
@@ -325,7 +325,11 @@ class NavigationContent extends ConsumerWidget {
                               .read(currentSectionProvider.notifier)
                               .setCurrentSection(app);
                           if (shouldPop) {
-                            Navigator.of(context).pop();
+                            // Use closeDrawer() (idempotent) rather than
+                            // Navigator.pop(); a rapid double tap with pop()
+                            // closes the drawer then pops the page route,
+                            // leaving a black screen.
+                            scaffoldGlobalKey.currentState?.closeDrawer();
```
teddy-chavez, line 332: Will this widget always be used as a drawer? Because we are deactivating pop for regular widgets right now, and I see that there is an isDrawer-boolean we can use :)

teddy-chavez: nvm, saw that this was already merged/released 😆

teddy-chavez (approved):

PR #2249, opened 2026-07-30
build(android): guard release APK against permission creep

AdamVe:
## Summary

A CameraX bump in 7.4.1 transitively pulled in `androidx.media3`, which
declares `ACCESS_NETWORK_STATE`, adding "view network connections" to the
Play Store permission listing. This removes the unused permission and adds a
build-time guard so future dependency bumps can't silently re-add
user-visible permissions.

## Changes

- Strip `ACCESS_NETWORK_STATE` from the merged manifest with
  `tools:node="remove"` — the app does no networking.
- Add `permission_guard.gradle.kts`: inspects the **release** merged manifest
  and fails `assembleRelease` if any permission is not on an explicit
  allow-list (`NFC`, `HIDE_OVERLAY_WINDOWS`, `CAMERA`, plus AndroidX's private
  `DYNAMIC_RECEIVER_NOT_EXPORTED_PERMISSION` self-permission).
- Guard is release-only; debug/profile builds keep Flutter's injected
  `INTERNET` permission for hot reload.

### DIFF
--- a/android/app/build.gradle
+++ b/android/app/build.gradle
@@ -89,6 +89,7 @@ kotlin {
 
 apply from: "signing.gradle"
 apply from: "collect_licenses.gradle"
+apply from: "permission_guard.gradle.kts"
 
 android.applicationVariants.all { variant ->
     variant.outputs.each { output ->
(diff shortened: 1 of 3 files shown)


teddy-chavez (approved):
Nice, good stuff!

